PR #6, opened 2016-01-13
Project activity widget

Prevole:
Not ready yet.


Prevole:
Ready for review

app/views/templates/project-recent-activity.html.slim
```diff
@@ -0,0 +1,35 @@
...
+    em No tests have been run recently.
+  .list-group ng-if='reports.length'
+    .list-group-item ng-repeat='report in reports'
+      span.singleRunner ng-if='report.runners.length == 1'
+        user-avatar ng-if='report.runners.length == 1' user='report.runners[0]'
+        strong.name {{ report.runners[0].name }}
+      span.multipleRunners ng-if='report.runners.length >= 2'
+        span.runner ng-repeat='runner in report.runners'
+          user-avatar user='runner' name-tooltip='true'
+      '  ran some tests
```
Prevole: I use `'` in place of `|` because of trailing space removed by my IDE. In fact, `'` add automatically a trailing space.

app/assets/javascripts/modules/project.js
```diff
@@ -30,4 +30,45 @@ angular.module('probedock.project', [ 'probedock.api', 'probedock.forms', 'probe
...
+        fetchReports();
+      }
+    });
+
+    function fetchReports() {
+      return api({
+        url: '/reports',
+        params: {
+          pageSize: 5,
+          organizationId: $scope.organization.id,
```
AlphaHydrae: The `organizationId` here is superfluous (but indeed required with the current version of the resource). By copying the following improvement to the reports API resource, you can remove it: https://github.com/probedock/probedock/blob/77bcad1e5ca75b86361412bd113e945a76646db4/app/api/probe_dock/metrics_api.rb#L31-L32

app/assets/javascripts/modules/project.js
```diff
@@ -30,4 +30,45 @@ angular.module('probedock.project', [ 'probedock.api', 'probedock.forms', 'probe
...
+          projectId: $scope.project.id,
+          withRunners: 1,
+          withProjects: 1,
+          withProjectVersions: 1,
+          withCategories: 1
```
AlphaHydrae: You should add the `withProjectCountsFor: $scope.project.id` option, that way you'll be able to retrieve the number of new tests specific to that project, not the report (which may contain multiple projects).

It could also be possible to display the passed/failed counts of the project instead of the report, but I would not do that change now, as it would then be inconsistent with the categories/tags, etc.

app/assets/javascripts/modules/project.js
AlphaHydrae: Also be sure that the loaded project is consistent with the current organization, as is done in this example: https://github.com/probedock/probedock/blob/77bcad1e5ca75b86361412bd113e945a76646db4/app/api/probe_dock/metrics_api.rb#L96

Prevole: Done

app/assets/javascripts/modules/project.js
Prevole: Done

PR #9, opened 2016-01-15
Test details page

Prevole:
Ready!


AlphaHydrae:
Links to tests from reports should conserve the status color (green/red/yellow) instead of the default link color.

app/assets/javascripts/modules/orgs.js
```diff
@@ -13,6 +13,22 @@ angular.module('probedock.orgs', [ 'probedock.api', 'probedock.auth', 'probedock
...
       },
 
+      getOrganization: function(orgId) {
+        var org =_.findWhere(service.organizations, { id: orgId });
+
+        if (_.isUndefined(org)) {
+          return api({
+            url: '/organizations/' + orgId
+          })
+          .then(function(response) {
```
AlphaHydrae: In general, the convention applied in the codebase so far is not to precede chained calls with new lines, except in special cases when there are many chained calls (e.g. states or module definitions).

app/policies/project_test_policy.rb
```diff
@@ -0,0 +1,50 @@
...
+# ProbeDock is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ProbeDock.  If not, see <http://www.gnu.org/licenses/>.
+class ProjectTestPolicy < ApplicationPolicy
+  def show?
+    user.try(:is?, :admin) || record.project.organization.try(:public?) || user.try('member_of?', record.project.try(:organization))
```
AlphaHydrae: Please use `admin?` instead of `user.try(:is?, :admin)`. Common checks are defined in [this helper class](https://github.com/probedock/probedock/blob/4d6281556b3efb7845647c1905d72830896ef9c9/app/policies/concerns/policy_helpers.rb#L50-L53).

While you're at it, you could add a `member_of?` helper to the class that takes the organization as argument. We'll replace assertions in other policy classes as we go forward.

Please use a symbol, not a string: `:member_of?` instead of `'member_of?'`.

Your last `try` call is either late or superfluous. If the organization was nil, the second `try` call in `record.project.organization.try(:public?)` would fail before reaching it. In the case of a show action for the project, the organization is guaranteed to be here anyway, so you could remove both of these `try` calls.

app/policies/project_test_policy.rb
```diff
@@ -0,0 +1,50 @@
...
+        json.name record.name
+        json.category record.description.category.name
+        json.key record.key.key if record.key.present?
+        json.resultsCount record.results_count
+        json.firstRunAt record.first_run_at.iso8601(3)
+        json.lastRunAt record.description.last_run_at.iso8601(3)
+        json.projectVersion record.description.project_version.name
+        json.passing record.description.passing
+        json.active record.description.active
+        json.tags record.description.tags.collect{ |t| t.name }
```
AlphaHydrae: FYI, simple attribute access blocks can be replaced with this syntax: `.collect(&:name)`. It is generally used across the codebase. But it's not important or required.

app/views/templates/report-details.html.slim
```diff
@@ -6,10 +6,10 @@
     .label.label-primary ng-if='!report' Loading...
     span ng-repeat='project in report.projects'
       span.label.label-primary ng-repeat='version in report.projectVersions | filter: {projectId: project.id}' {{ project | projectName }} {{ version.name }}
-      |  
+      |
```
AlphaHydrae: Replace these `|` by `'` to make sure the extra spaces remain (2 other occurrences below).

app/views/templates/report-details.html.slim
```diff
@@ -86,7 +86,8 @@
       .panel id='{{ testAnchor(result) }}' ng-class='{"panel-success": result.passed && result.active, "panel-danger": !result.passed && result.active, "panel-warning": !result.active}'
         .panel-body
           test-key-label.pull-right ng-if='result.key' key='result.key'
-          h4 ng-class='{"text-success": result.passed && result.active, "text-danger": !result.passed && result.active, "text-warning": !result.active}' {{ result.name }}
+          h4 ng-class='{"text-success": result.passed && result.active, "text-danger": !result.passed && result.active, "text-warning": !result.active}'
+            a href='#' ui-sref='org.tests.show({ testId: result.testId, orgId: result.project.organizationId })' {{ result.name }}
```
AlphaHydrae: Shouldn't the `ordId` parameter be `orgName`?

app/views/templates/test.html.slim
```diff
@@ -0,0 +1,42 @@
...
+              dd
+                strong Last status
+              dt.test-details-status
+                span.glyphicon.glyphicon-thumbs-up.test-details-passed ng-if='test.passing && test.active' tooltip='Passed' tooltip-append-to-body='true'
+                span.glyphicon.glyphicon-thumbs-down.test-details-failed ng-if='!test.passing && test.active' tooltip='Failed' tooltip-append-to-body='true'
+                span.glyphicon.glyphicon-ban-circle.test-details-inactive ng-if='!test.active' tooltip='Inactive' tooltip-append-to-body='true'
+              dd
+                strong Contributors
+              dd
+                user-avatar ng-repeat='contribution in test.contributions' user='contribution.user' tooltip='{{ contribution.user.name }}' tooltip-append-to-body='true'
```
AlphaHydrae: In the current implementation, the list of contributors can be empty for some tests (and that is the case for some of the current tests). You should add a conditional `.empty` text to allow for the possibility.

app/assets/javascripts/modules/orgs.js
Prevole: Like this ?

```
          return api({
            url: '/organizations/' + orgId
          }).then(function(response) {
            service.addOrganization(response.data);
            return response.data;
          });
```
In my opinion, adding one return line is better :)

app/policies/project_test_policy.rb
Prevole: Copied from this:
```
class ProjectPolicy < ApplicationPolicy
  def create?
    user.is?(:admin) || user.membership_in(record.organization).try(:is?, :admin)
  end

  def index?
    user.try(:is?, :admin) || organization.try(:public?) || user.try(:member_of?, organization)
  end

  def show?
    user.try(:is?, :admin) || organization.try(:public?) || user.try(:member_of?, organization)
  end

  def update?
    user.is?(:admin) || user.membership_in(record.organization).try(:is?, :admin)
  end
```
And later, fixed with your help for exception raised around this `try` stuff.

features/project_test.feature
```diff
@@ -0,0 +1,286 @@
...
+    Given private organization Rebel Alliance exists
+    And user hsolo who is a member of Rebel Alliance exists
+    And user lskywalker who is a member of Rebel Alliance exists
+    And user wantilles who is a member of Rebel Alliance exists
+    And project X-Wing exists within organization Rebel Alliance
+    And test "Ion engine should provide thrust" was created by lskywalker with key k1 for version 1.0.0 of project X-Wing
+    And test "Ion engine should provide thrust" has category C1 and tags tag1, tag2 and tickets ticket1, ticket2 for version 1.0.0
+    And test "Blasters should fire" was first run by lskywalker for version 1.0.0 of project X-Wing
+    And test "Blasters should fire" has category C2 for version 1.0.0
+    When hsolo sends a GET request to /api/tests/{@idOf: Ion engine should provide thrust}?organizationId={@idOf: Rebel Alliance}
```
AlphaHydrae: Why is there an `organizationId` query param in this GET request? Same question for the other tests (sometimes with `projectId`)? It should not be necessary to add these parameters for a GET on a single resource.

features/project_test.feature
```diff
@@ -0,0 +1,286 @@
...
+  - number of results
+  - first run date
+  - last run date
+  - latest project version
+  - passing (last result)
+  - active/inactive
+
+
+
+  Scenario: An organization member should be able to get the test details without project data in his organization
```
AlphaHydrae: In the description of the default case, I would not mention was is not there (i.e. "without project data"). Otherwise you would also need to add "without contributions" and all other absent filters that we may add.

app/policies/project_test_policy.rb
Prevole: Here we go with two new helpers:
```
  # Indicates whether the current user is member of the given organization
  def member_of?(organization)
    user.kind_of(User) && organization.kind_of?(Organization) && user.try(:member_of?, organization)
  end

  # Indicates whether the given organization is public
  def public?(organization)
    organization.kind_of?(Organization) && organization.try(:public?)
  end
```

features/project_test.feature
```diff
@@ -0,0 +1,286 @@
+@api @test
+Feature: Test
+
+  Users should be able to retrieve the test details by its id.
```
AlphaHydrae: Prefer `a test`, or `the details of a test`, or `test details` to `the test details` (which sounds weird).

app/assets/javascripts/modules/orgs.js
AlphaHydrae: I prefer it that way, but I don't care one way or another, as long as we define a clear convention and apply it to all new and updated code.

app/views/templates/test.html.slim
Prevole: Approved?
```
              dd
                user-avatar ng-if='test.contributions.length > 0' ng-repeat='contribution in test.contributions' user='contribution.user' tooltip='{{ contribution.user.name }}' tooltip-append-to-body='true'
                p.empty ng-if='test.contributions.length === 0'
                  em No users have been contributed yet.
```

features/project_test.feature
Prevole: No idea. Removed, cucumber is ok :)

features/project_test.feature
Prevole: Fixed

features/project_test.feature
Prevole: Fixed

app/assets/javascripts/modules/orgs.js
Prevole: I Fixed it in your way but be prepared to do the remark again an undefined number of times as I will continue to use my convention in other projects :)

Prevole: Is that ok ?
```
    api({
      url: '/tests/' + $stateParams.testId,
      params: {
        organizationId: $stateParams.orgId,
        withProject: 1,
        withContributions: 1
      }
    }).then(function(response) {
      $scope.test = response.data;

      orgs
        .getOrganization(response.data.project.organizationId)
        .then(function(organization) {
          $scope.organization = organization;
        });
    });
```

app/policies/project_test_policy.rb
Prevole: Fixed

app/views/templates/report-details.html.slim
Prevole: Fixed

app/views/templates/report-details.html.slim
Prevole: In fact, I was confused by the GET /...?organizationId where the orgId is required. But without this query param, there is no need to the parameter in the state anymore.

Finally, `orgId` or `orgName` is not useful if we use `currentOrganization` to retrieve the organization name. The name is used to show <organization> > <project> under the test name which is the title of the page.

If it is ok for you, we can go with only test id:
```
a href='#' ui-sref='org.tests.show({ testId: result.testId })' {{ result.name }}
```
which result in simplification of the `TestCtrl`:
```
  .controller('TestCtrl', function (api, forms, orgs, projects, $scope, $state, $stateParams) {
    orgs.forwardData($scope);

    api({
      url: '/tests/' + $stateParams.testId,
      params: {
        withProject: 1,
        withContributions: 1
      }
    }).then(function(response) {
      $scope.test = response.data;
    });
  })
```
Before, there was a REST call to retrieve the organization data.

And the template is also updated to show the content only when the test has been loaded:
```
    h4 ng-if='!test' Loading...
    h4 ng-if='test' {{ currentOrganization | orgName }} > {{ test.project | projectName }}
```

app/assets/javascripts/modules/orgs.js
AlphaHydrae: Approved. The same convention applies to `else if` and `else`, e.g.:

```
if (cond) {

} elseif (cond) {

} else {

}
```

app/assets/javascripts/modules/test.js
```diff
@@ -0,0 +1,25 @@
...
+    orgs.forwardData($scope);
+
+    api({
+      url: '/tests/' + $stateParams.testId,
+      params: {
+        organizationId: $stateParams.orgId,
+        withProject: 1,
+        withContributions: 1
+      }
+    })
```
AlphaHydrae: Remove new line.

app/policies/project_test_policy.rb
AlphaHydrae: Good. I know the existing policies don't all use the helpers because I haven't yet taken the time to refactor all of them.

app/views/templates/report-details.html.slim
AlphaHydrae: `ui-sref='org.tests.show({ testId: result.testId })'`

Does this work without the `orgName` parameter? I assume it's because it keeps the same value by default, because the parameter is required for a child of the `org.` state.

app/views/templates/report-details.html.slim
```diff
@@ -104,7 +105,7 @@
               |  {{ result.duration | durationFormat }}
           pre ng-if='result.message' {{ result.message }}
     p.empty
-      span ng-if='noMoreResults' All results have been loaded. 
+      span ng-if='noMoreResults' All results have been loaded.
```
AlphaHydrae: Add a `'` between the `span` and the `em` to ensure there is at least one space in the case where the two messages are displayed at the same time.

app/views/templates/test.html.slim
AlphaHydrae: Replace the message by "No users are known to have contributed to this test."

features/step_definitions/test_steps.rb
```diff
@@ -45,8 +46,22 @@
   organization = project_version.project.organization
 
   description.category = Category.where(organization_id: organization.id, name: category_name).first_or_create if category_name
-  raise 'not yet implemented' if tag_names
-  raise 'not yet implemented' if ticket_names
+
+  if tag_names
+    tag_names.split(',').each do |t|
+      tag = create :tag, name: t, organization: organization
```
AlphaHydrae: It would be better to use `Tag.where(organization: organization, name: t).first_or_create` so that the step doesn't fail if used for different tests with the same tag names. Same remark for the tickets.

app/views/templates/test.html.slim
Prevole: Fixed

app/assets/javascripts/modules/test.js
Prevole: Fixed

app/views/templates/report-details.html.slim
Prevole: Fixed

features/step_definitions/test_steps.rb
Prevole: Fixed

app/views/templates/report-details.html.slim
Prevole: Fixed with `orgName` used as state param.

app/policies/project_test_policy.rb
Prevole: Fixed

features/project_test.feature
```diff
@@ -0,0 +1,286 @@
...
+        "passing": true,
+        "active": true,
+        "tags": [ "tag1", "tag2" ],
+        "tickets": [ "ticket1", "ticket2" ]
+      }
+      """
+    And nothing should have been added or deleted
+
+
+@wip
```
AlphaHydrae: Unwip.

Prevole: Unwiped

PR #10, opened 2016-01-20
Test evolution widget

Prevole:
Ready for review

### DIFF
--- a/app/api/probe_dock/metrics_api.rb
+++ b/app/api/probe_dock/metrics_api.rb
@@ -18,7 +18,7 @@
 module ProbeDock
   class MetricsApi < Grape::API
     DEFAULT_NB_DAYS_FOR_REPORTS = 30
-    DEFAULT_MAX_NB_DAYS_FOR_REPORTS = 120
+    MAX_NB_DAYS_FOR_REPORTS = 120
 
     namespace :metrics do
       before do
@@ -78,7 +78,7 @@ module ProbeDock
         authorize! :organization, :data
 
         nb_days = params[:nbDays].to_i
-        nb_days = DEFAULT_MAX_NB_DAYS_FOR_REPORTS if nb_days > DEFAULT_MAX_NB_DAYS_FOR_REPORTS
+        nb_days = MAX_NB_DAYS_FOR_REPORTS if nb_days > MAX_NB_DAYS_FOR_REPORTS
         nb_days = DEFAULT_NB_DAYS_FOR_REPORTS if nb_days <= 0
 
         start_date = (nb_days - 1).days.ago.beginning_of_day
@@ -91,11 +91,11 @@ module ProbeDock
         filter_by_users = params[:userIds].present? && params[:userIds].kind_of?(Array)
 
         # Prepare additional join for test_payloads
-        more_joins = []
-        more_joins << { project_version: :project } if filter_by_projects
-        more_joins << :runner if filter_by_users
+        payload_joins = []
+        payload_joins << { project_version: :project } if filter_by_projects
+        payload_joins << :runner if filter_by_users
 
-        rel = rel.joins(test_payloads: more_joins) unless more_joins.empty?
+        rel = rel.joins(test_payloads: payload_joins) unless payload_joins.empty?
 
         # Apply filters
         rel = rel.where 'projects.api_id IN (?)', params[:projectIds].collect(&:to_s) if filter_by_projects
--- a/app/api/probe_dock/metrics_api.rb
+++ b/app/api/probe_dock/metrics_api.rb
@@ -17,6 +17,9 @@
 # along with ProbeDock.  If not, see <http://www.gnu.org/licenses/>.
 module ProbeDock
   class MetricsApi < Grape::API
+    DEFAULT_NB_DAYS_FOR_REPORTS = 30
+    DEFAULT_MAX_NB_DAYS_FOR_REPORTS = 120
+
     namespace :metrics do
       before do
         authenticate
@@ -34,7 +37,7 @@ module ProbeDock
         end
       end
 
-      get :newTests do
+      get :newTestsByDay do
         authorize! :organization, :data
 
         start_date = 29.days.ago.beginning_of_day
@@ -65,6 +68,49 @@ module ProbeDock
         30.times do |i|
           date = current_date.strftime('%Y-%m-%d')
           result << { date: date, testsCount: counts.fetch(date, 0) }
+          current_date += 1.day
+        end
+
+        result
+      end
+
+      get :reportsByDay do
+        authorize! :organization, :data
+
+        nb_days = params[:nbDays].to_i
+        nb_days = DEFAULT_MAX_NB_DAYS_FOR_REPORTS if nb_days > DEFAULT_MAX_NB_DAYS_FOR_REPORTS
+        nb_days = DEFAULT_NB_DAYS_FOR_REPORTS if nb_days <= 0
+
+        start_date = (nb_days - 1).days.ago.beginning_of_day
+
+        rel = TestReport.where('test_reports.organization_id = ?', current_organization.id).select("count(distinct test_reports.id) as runs_count, date_trunc('day', test_reports.ended_at) as runs_day")
+        rel = rel.where 'test_reports.ended_at >= ?', start_date
+
+        # Set filters by
+        filter_by_projects = params[:projectIds].present? && params[:projectIds].kind_of?(Array)
+        filter_by_users = params[:userIds].present? && params[:userIds].kind_of?(Array)
+
+        # Prepare additional join for test_payloads
+        more_joins = []
+        more_joins << { project_version: :project } if filter_by_projects
+        more_joins << :runner if filter_by_users
+
+        rel = rel.joins(test_payloads: more_joins) unless more_joins.empty?
+
+        # Apply filters
+        rel = rel.where 'projects.api_id IN (?)', params[:projectIds].collect(&:to_s) if filter_by_projects
+        rel = rel.where 'users.api_id IN (?)', params[:userIds].collect(&:to_s) if filter_by_users
+
+        rel = rel.group('runs_day').order('runs_day DESC')
+
+        counts = rel.to_a.inject({}){ |memo,data| memo[data.runs_day.strftime('%Y-%m-%d')] = data.runs_count; memo }
+
+        result = []
+        current_date = start_date
+
+        nb_days.times do |i|
+          date = current_date.strftime('%Y-%m-%d')
+          result << { date: date, runsCount: counts.fetch(date, 0) }
           current_date += 1.day
         end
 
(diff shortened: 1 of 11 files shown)


app/api/probe_dock/metrics_api.rb
```diff
@@ -17,6 +17,9 @@
 # along with ProbeDock.  If not, see <http://www.gnu.org/licenses/>.
 module ProbeDock
   class MetricsApi < Grape::API
+    DEFAULT_NB_DAYS_FOR_REPORTS = 30
+    DEFAULT_MAX_NB_DAYS_FOR_REPORTS = 120
```
AlphaHydrae: I would remove `DEFAULT_` from the second variable's name. The first one is the default, the second one is the max, not the "default max", whatever that means.

app/api/probe_dock/metrics_api.rb
```diff
@@ -71,6 +74,49 @@ def current_organization
...
+
+        rel = TestReport.where('test_reports.organization_id = ?', current_organization.id).select("count(distinct test_reports.id) as runs_count, date_trunc('day', test_reports.ended_at) as runs_day")
+        rel = rel.where 'test_reports.ended_at >= ?', start_date
+
+        # Set filters by
+        filter_by_projects = params[:projectIds].present? && params[:projectIds].kind_of?(Array)
+        filter_by_users = params[:userIds].present? && params[:userIds].kind_of?(Array)
+
+        # Prepare additional join for test_payloads
+        more_joins = []
```
AlphaHydrae: I would call this variable `payload_joins` to make it clearer.

app/assets/javascripts/modules/dashboard.js
```diff
@@ -52,23 +52,42 @@ angular.module('probedock.dashboard', [ 'probedock.api', 'probedock.orgs', 'prob
       controller: 'NewTestsLineChartCtrl',
       templateUrl: '/templates/new-tests-line-chart.html',
       scope: {
-        organization: '='
+        organization: '=',
+        projectId: '='
```
AlphaHydrae: Why pass `projectId` and not `project` here? This seems to differ from most (if not all) other directives.

app/assets/javascripts/modules/dashboard.js
```diff
@@ -52,23 +52,42 @@ angular.module('probedock.dashboard', [ 'probedock.api', 'probedock.orgs', 'prob
...
+    };
+
     $scope.chart = {
       data: [],
       labels: [],
       params: {
         projectIds: [],
-        userIds: []
+        userIds: [],
+        type: 'written'
```
AlphaHydrae: Putting the type in the `chart` object rather than `chart.params` would be simpler and avoid the `_.omit` call in `fetchMetrics` below.

app/assets/javascripts/modules/dashboard.js
```diff
@@ -52,23 +52,42 @@ angular.module('probedock.dashboard', [ 'probedock.api', 'probedock.orgs', 'prob
...
       },
       options: {
         pointHitDetectionRadius: 5,
-        tooltipTemplate: '<%= value %> new tests on <%= label %>'
+        // Will be set from chartConfig based on chart type
+        tooltipTemplate: '',
+        scaleLabel: function(object) {
+          return '  ' + object.value;
```
AlphaHydrae: Please add a comment for this fix, if possible with the link to the source of the solution.

app/assets/javascripts/modules/dashboard.js
```diff
@@ -130,15 +156,19 @@ angular.module('probedock.dashboard', [ 'probedock.api', 'probedock.orgs', 'prob
         return;
       }
 
+      $scope.chart.options.tooltipTemplate = chartConfig[$scope.chart.params.type].tooltipTemplate;
```
AlphaHydrae: It seems a bit dirty to put this code in a function called `fetchMetrics`. I would create another `updateChart` function that is also called in the same circumstances. The first 3 lines of the function could also go in there.

app/views/templates/new-tests-line-chart.html.slim
```diff
@@ -1,23 +1,38 @@
...
+                span ng-bind-html='project | projectName'
+        col-xs-12 ng-class='{"col-sm-4": !projectId, "col-sm-6": projectId }'
+          .form-group
+            label for='new-tests-line-chart-users' Filter by user
+            ui-select#new-tests-line-chart-users ng-model='chart.params.userIds' multiple=true reset-search-input=true
+              ui-select-match placeholder='All users' {{ $item.name }}
+              ui-select-choices repeat='user.id as user in userChoices | filter: $select.search'
+                span ng-bind-html='user.name'
+      .empty ng-if='chart.data.length === 0'
+        em No tests have been added recently.
```
AlphaHydrae: Change this text to "No recent activity."

features/reports_by_day_metrics.feature
```diff
@@ -0,0 +1,207 @@
...
+  - runsCount for each day
+  - date for each day
+
+
+  Background:
+    Given private organization Rebel Alliance exists
+    And user hsolo who is a member of Rebel Alliance exists
+    And user lskywalker who is a member of Rebel Alliance exists
+    And project X-Wing exists within organization Rebel Alliance
+    And test result report report 1 was generated 2 days ago for organization Rebel Alliance
```
AlphaHydrae: "test result report report 1" reads very weird. I would prefer replacing "report 1" by "R1" and similarly for other such occurrences.

features/step_definitions/metrics_steps.rb
```diff
@@ -0,0 +1,3 @@
+Given /^assuming (\d+) days is set to retrieve the reports by day$/ do |nb_days|
```
AlphaHydrae: That doesn't mean anything. Replace with "assuming reports by day metrics are calculated for X days by default".

features/reports_by_day_metrics.feature
```diff
@@ -0,0 +1,207 @@
...
+    And test payload payload 1 sent by lskywalker for version 1.0.0 of project X-Wing was used to generate report report 1
+    And test payload payload 2 sent by hsolo for version 1.0.0 of project X-Wing was used to generate report report 1
+    And test result report report 2 was generated 2 days ago for organization Rebel Alliance
+    And test payload payload 3 sent by lskywalker for version 1.0.0 of project X-Wing was used to generate report report 2
+    And test result report report 3 was generated 1 days ago for organization Rebel Alliance
```
AlphaHydrae: As the regexp is defined with an optional plural for "days", use "day" for 1.

features/step_definitions/report_steps.rb
```diff
@@ -1,9 +1,18 @@
...
   options = {
     uid: uid,
     organization: named_record(organization_name)
   }
 
+  if days
+    date = days.to_i.days.ago
+
+    options.merge!({
+      started_at: date - 3.minutes,
```
AlphaHydrae: I would use `date` as is for `started_at` and add some time for `ended_at` instead of subtracting for both. Agreed, it's unlikely, but technically this could end up not being the same start & end day, whereas if you only add to `ended_at`, at least the start date will always correspond to what is written in the scenario.

spec/support/api_spec_helper.rb
```diff
@@ -78,6 +78,13 @@ def interpolate_string value, options = {}
       named_record(m[1]).api_id
     elsif m = value.match(/^@valueOf\((.*), (.*)\)$/)
       named_record(m[1]).send(m[2])
+    elsif m = value.match(/^@date\((.*)\)$/)
+      if m[1].match(/(today|now)/)
+        Time.now.strftime('%Y-%m-%d')
+      else
+        m = value.match(/(\d+) (second|minute|hour|day|week|month|year)s? ago/)
```
AlphaHydrae: This line should be combined with the `else` into:

```
elsif m = value.match(/(\d+) (second|minute|hour|day|week|month|year)s? ago/)
```

And you should add an `else` that does something like:

```
else
  raise "Unknown date expectation format"
end
```

app/api/probe_dock/metrics_api.rb
Prevole: Fixed

app/api/probe_dock/metrics_api.rb
Prevole: Fixed

app/assets/javascripts/modules/dashboard.js
Prevole: Fixed (also in template where the directive is used)

app/assets/javascripts/modules/dashboard.js
Prevole: Fixed (also in template where the controller is used)

app/assets/javascripts/modules/dashboard.js
Prevole: Comment added with two links.

app/assets/javascripts/modules/dashboard.js
Prevole: Not sure if your remark is applicable since the function is called `showMetrics` and not `fetchMetrics`. Can you review this again please.

app/views/templates/new-tests-line-chart.html.slim
Prevole: Fixed

features/reports_by_day_metrics.feature
Prevole: Fixed

features/step_definitions/metrics_steps.rb
Prevole: Fixed (also the feature using this step)

features/reports_by_day_metrics.feature
Prevole: I will go further by adding more constraints in the regex like that:
```
(?: (1 day|[2-9]\d* days) ago)?
```

app/assets/javascripts/modules/dashboard.js
AlphaHydrae: Ignore the remark. It's the code fold's fault.

features/step_definitions/report_steps.rb
Prevole: Should I change also there ? https://github.com/probedock/probedock/blob/master/spec/factories/test_report.rb#L21-L22

features/reports_by_day_metrics.feature
AlphaHydrae: I wouldn't constraint the regexp, as it's quite un-Rails-like (e.g. Rails accepts `1.days.ago` too), and the simpler the regexp the better.

features/step_definitions/report_steps.rb
AlphaHydrae: Go ahead.

features/reports_by_day_metrics.feature
AlphaHydrae: See the comment I left on a similar regexp in one of the other PRs.

spec/support/api_spec_helper.rb
Prevole: Improved

features/step_definitions/report_steps.rb
Prevole: Fixed

PR #13, opened 2016-01-21
Project health widget

Prevole:
Not yet ready but review on the API side is welcome.

### DIFF
--- a/app/api/probe_dock/metrics_api.rb
+++ b/app/api/probe_dock/metrics_api.rb
@@ -33,6 +33,8 @@ module ProbeDock
             Organization.active.where(normalized_name: params[:organizationName].to_s.downcase).first!
           elsif params[:projectId].present?
             Organization.active.joins(:projects).where('projects.api_id = ?', params[:projectId].to_s).first!
+          elsif params[:projectVersionId].present?
+            Organization.active.joins(:projects => [:versions]).where('project_versions.api_id = ?', params[:projectVersionId]).first!
           end
         end
       end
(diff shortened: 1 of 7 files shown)


app/api/probe_dock/metrics_api.rb
```diff
@@ -117,6 +119,40 @@ def current_organization
...
+        # Retrieve the data from the specified version otherwise take the most recent project version to retrieve the data
+        if params[:projectVersionId].present?
+          rel = rel.joins(:project_version).where('project_versions.api_id = ?', params[:projectVersionId])
+        else
+          project_version = ProjectVersion.joins(:project).where('projects.api_id = ?', params[:projectId]).order('created_at DESC').first
+          rel = rel.where('project_versions.id = ?', project_version.id)
+        end
+
+        # Calculate the different metrics
+        result = rel.to_a.inject({testsCount: 0, passedTestsCount: 0, inactiveTestsCount: 0, inactivePassedTestsCount: 0}) do |memo, data|
```
AlphaHydrae: Iterating over descriptions will be quite slow for projects with thousands of tests. You could batch the queries, but I think it would be better to have the database count the items.

app/api/probe_dock/metrics_api.rb
```diff
@@ -117,6 +119,40 @@ def current_organization
...
+          memo[:testsCount] += 1
```
AlphaHydrae: `testsCount` should be the total number of tests, meaning you should retrieve it with a query on `ProjectTest`, otherwise it doesn't take unused tests into account.

AlphaHydrae: Also don't forget to calculate and add the `unusedTestsCount`, which you should be able to calculate with: the number of tests (total) minus the number of test descriptions (the ones that have been run).

app/api/probe_dock/metrics_api.rb
```diff
@@ -117,6 +119,40 @@ def current_organization
...
+          memo[:passedTestsCount] += data.passing ? 1 : 0
+          memo[:inactiveTestsCount] += !data.active ? 1 : 0
+          memo[:inactivePassedTestsCount] += data.passing && !data.active ? 1 : 0
+          memo
+        end
+
+        # Enrich with project version data
+        result.merge!({
+          projectVersion: {
```
AlphaHydrae: I would load the project version in all cases and use the policy to serialize it here.

Prevole: Fixed

app/api/probe_dock/metrics_api.rb
Prevole: Fixed

app/api/probe_dock/metrics_api.rb
Prevole: Fixed

PR #18, opened 2016-01-26
Project test payload dropzone widget

AlphaHydrae:


app/assets/javascripts/modules/project.js
```diff
@@ -70,6 +68,199 @@ angular.module('probedock.project', [ 'probedock.api', 'probedock.forms', 'probe
...
+      }).then(function(res) {
+        $scope.categories = res.data;
+
+        // if there is no existing category, automatically switch
+        // to the free input field to create a new one
+        $scope.formConfig.newCategory = !res.data.length;
+      });
+    }
+
+    function fetchVersions() {
```
Prevole: There is a directive to select a version. Is it possible to re-use/adapt it ?

app/policies/concerns/policy_helpers.rb
```diff
@@ -63,12 +63,12 @@ def technical?
   end
 
   # Indicates whether the current user is member of the given organization
-  def member_of?(organization)
+  def member_of? organization
```
Prevole: My prefered IDE does not like definitions of methods without parenthesis and the help will refer to this: https://github.com/bbatsov/ruby-style-guide#syntax

AlphaHydrae: That's not the convention that's been applied in the codebase so far. I don't mind changing the convention as long as you agree that we'll refactor all existing code as we update it, and that we'll point out all missing parentheses in future reviews.

Prevole: Fine for me

PR #19, opened 2016-01-26
Attempt to extract the various selects in its their directives

Prevole:
Just an attempt to extract the user/project choice control in their own directives. Only done for user choice.


app/assets/javascripts/boot.js.erb
```diff
@@ -55,6 +55,7 @@ angular.module('probedock', [
   'probedock.tables',
   'probedock.testActivityWidget',
   'probedock.testsByWeekWidget',
+  'probedock.userChoiceCtrl',
```
AlphaHydrae: Calling this `somethingCtrl` is asking for confusion with controllers. Please rename it `userChoiceSelect` or something else.

app/assets/javascripts/modules/userChoiceCtrl.js
```diff
@@ -0,0 +1,43 @@
+angular.module('probedock.userChoiceCtrl', [ 'probedock.api' ])
+  .directive('userChoiceCtrl', function() {
+    return {
+      restrict: 'E',
+      controller: 'UserChoiceCtrl',
+      templateUrl: '/templates/user-choice-ctrl.html',
+      scope: {
+        organization: '=',
+        update: '='
```
AlphaHydrae: There's a dedicated way of passing functions to directives which reduces the coupling with the parent scope. Use `&` instead of `=` and call it `onUpdate` (convention).

app/assets/javascripts/modules/userChoiceCtrl.js
```diff
@@ -0,0 +1,43 @@
...
+
+    $scope.$watch('organization', function(value) {
+      if (value) {
+        fetchUserChoices();
+      }
+    });
+
+    $scope.$watch('model.userIds', function(value) {
+      if (value && $scope.update) {
+        $scope.update(value);
```
AlphaHydrae: Replace with `$scope.onUpdate({ userIds: value });` to make it work with `&`.

app/views/templates/tests-by-week-widget.html.slim
```diff
@@ -17,11 +17,6 @@
               ui-select-choices repeat='project.id as project in projectChoices | filter: $select.search'
                 span ng-bind-html='project | projectName'
         .col-xs-12 ng-class='{"col-sm-6": !project }'
-          .form-group
-            label for='new-tests-line-chart-users' Filter by user
-            ui-select#new-tests-line-chart-users ng-model='chart.params.userIds' multiple=true reset-search-input=true
-              ui-select-match placeholder='All users' {{ $item.name }}
-              ui-select-choices repeat='user.id as user in userChoices | filter: $select.search'
-                span ng-bind-html='user.name'
+          user-choice--ctrl project='project' organization='organization' update='updateUserIds'
```
AlphaHydrae: Replace with `on-update='updateUserIds(userIds)'` to make it work with `&`.

app/assets/javascripts/modules/userChoiceSelect.js
```diff
@@ -0,0 +1,41 @@
+angular.module('probedock.userChoiceSelect', [ 'probedock.api' ])
+  .directive('userChoiceSelect', function() {
+    return {
+      restrict: 'E',
+      controller: 'UserChoiceSelectCtrl',
+      templateUrl: '/templates/user-choice-select.html',
+      scope: {
+        organization: '=',
+        userIds: '=ngModel'
```
Prevole: tried: `'model.userIds': '=ngModel'` but it does not work. Therefore, I use an alias with `userIds` and a `$watch` to update it from `model.userIds` which is used by ui-select.

PR #24, opened 2016-02-02
Added tests by categories widget

Prevole:
Pre-review


Prevole:
- Added a new JS lib with a corresponding directive to open/close kind of popus for presenting filters.
- Modified drop-ng.js in vendor assets to fix a problem of compatibility with ui-select. A PR is open there to have the fix in a release: https://github.com/stevenh77/drop-ng/pull/15
- Waiting for comments

Prevole:
- Added the table with the remaining categories. The table is shown after a click on a button.

app/api/probe_dock/metrics_api.rb
```diff
@@ -145,13 +168,7 @@ def count_new_tests_by_interval(start_date, interval)
...
         authorize! :organization, :data
 
-
-        # Retrieve the data from the specified version otherwise take the most recent project version to retrieve the data
-        if params[:projectVersionId].present?
-          project_version = ProjectVersion.where('api_id = ?', params[:projectVersionId]).first
-        else
-          project_version = ProjectVersion.joins(:project).where('projects.api_id = ?', params[:projectId]).order('created_at DESC').first
-        end
+        project_version = project_version()
```
AlphaHydrae: I would avoid this, it looks very confusing. It's a good idea in general to avoid overriding methods with local variables. (And on a side note, in this case it is not compliant with the style guide as there should be no parentheses for a method call with no arguments.) We could name the method `current_project_version`, as it takes the project version defined by the current params like `current_organization` does. And leave the local variable as is.

app/assets/javascripts/modules/testsByCategoryWidget.js
```diff
@@ -0,0 +1,166 @@
+angular.module('probedock.testsByCategoryWidget', [ 'probedock.api' ])
+  .directive('testsByCategoryWidget', function() {
+    return {
+      restrict: 'E',
+      controller: 'TestsByCategoryBarsCtrl',
+      templateUrl: '/templates/tests-by-category-widget.html',
+      scope: {
+        organization: '=',
+        project: '=',
+        nbCategories: '=?'
```
AlphaHydrae: Small thing: in the case of the by day/week widgets, the number you pass to the directive is always the number of days/weeks that will be displayed, so it makes sense to have `nbDays` as a parameter name. But in this case, there may be fewer categories displayed, so it's a maximum. I would call this `maxNbCategories`. And maybe add a comment (on the directive or on the scope parameter) to explain that an "Other" will be displayed and whether it is included in this number.

app/assets/stylesheets/modules/drop-ng.css.styl
```diff
@@ -0,0 +1,6 @@
+.drop-element
+  visibility  hidden
```
AlphaHydrae: Extra space.

app/assets/javascripts/modules/testsByCategoryWidget.js
```diff
@@ -0,0 +1,166 @@
...
+            testsByCategory.push({
+              name: 'Others',
+              testsCount: others,
+              percent: 100 / total * others,
+              others: othersArray
+            });
+          }
+
+        // As there is only one other category, just show the category in place of others
+        } else {
```
AlphaHydrae: Rather than write:

```
if (!_.isEmpty(otherCategories)) {
  if (otherCategories.length > 1) {
  } else {
  }
}
```

Couldn't we write?

```
if (otherCategories.length > 1) {
} else if (otherCategories.length == 1) {
}
```

app/views/templates/tests-by-category-widget.html.slim
```diff
@@ -0,0 +1,66 @@
+.panel.panel-default
+  .panel-heading
+    .row
+      .col-md-6
+        h3.panel-title Tests by category
+      .col-md-6
+        div style='float:right;'
+          a.btn.btn-default.btn-xs href='#' role='button'
+            i.glyphicon.glyphicon-filter
+            drop constrain-to-scroll-parent='true' constrain-to-window='true' open-on='"click"' position='"left top"' tether-options='{offset: "-5px -5px"}' before-close='beforeClose()'
```
AlphaHydrae: When I tried it, this button would open the dropdown, but not close it when I clicked on it again. It should.

app/views/templates/tests-by-category-widget.html.slim
```diff
@@ -0,0 +1,66 @@
+.panel.panel-default
+  .panel-heading
+    .row
+      .col-md-6
+        h3.panel-title Tests by category
+      .col-md-6
+        div style='float:right;'
```
AlphaHydrae: Use `.pull-right` from Bootstrap (if it works as intended).

app/views/templates/tests-by-category-widget.html.slim
```diff
@@ -0,0 +1,66 @@
...
+          a.btn.btn-default.btn-xs href='#' role='button'
```
AlphaHydrae: Add a "Filters" tooltip to this button.

app/views/templates/tests-by-category-widget.html.slim
```diff
@@ -0,0 +1,66 @@
...
+      em Loading...
+  .panel-body ng-if='testsByCategory.length > 0'
+    table.table.tests-by-category-table
+      tr
+        td.tests-by-category-td ng-repeat='record in testsByCategory'
+          span style='border-left-color: {{ getColor($index) }};'
+            strong {{ record.name }}
+            |  {{ record.percent | number : 1 }}% / {{ record.testsCount }}
+            ng-pluralize count='record.testsCount' when='{"1": " test", "other": " tests"}'
+            a.tests-by-category-show-more.btn.btn-default.btn-xs href='#' role='button' ng-if='record.others'
```
AlphaHydrae: Add a "Show all other categories" tooltip to this button.

features/tests_by_category_metrics.feature
```diff
@@ -0,0 +1,290 @@
...
+  - project
+  - user
+  - project and user
+  - project version and user
+
+  The following result is provided:
+  - name of the category
+  - the number of tests for this category
+
+  In some cases, tests do not have a category. In this case, a dedicated property is available to get
```
AlphaHydrae: This description lies. The `noCategoryTestsCount` property is not available only "in this case". It is always returned by the API, whether there are tests with no category or not.

features/tests_by_category_metrics.feature
```diff
@@ -0,0 +1,290 @@
...
+            "testsCount": 1
+          }
+        ]
+      }
+      """
+    And nothing should have been added or deleted
+
+
+
+  Scenario: A user should be able to retrieve the number of tests by categories from public organization
```
AlphaHydrae: "A member from another organization should be able to retrieve the number of tests by categories from a public organization" would be clearer.

features/tests_by_category_metrics.feature
```diff
@@ -0,0 +1,290 @@
...
+            "testsCount": 1
+          }
+        ]
+      }
+      """
+    And nothing should have been added or deleted
+
+
+
+  Scenario: An anonymous user should be able to retrieve the number of tests by categories from public organization
```
AlphaHydrae: "from a public organization"

app/api/probe_dock/metrics_api.rb
```diff
@@ -62,6 +68,12 @@ def apply_projects_and_users_filters(rel_for_filtering)
           rel
         end
 
+        def apply_projects_and_users_filters(rel_for_filtering)
+          rel = apply_projects_filters rel_for_filtering
```
AlphaHydrae: In previous methods, you have defined a `rel_for_filtering` argument and the first thing you do is re-declare a `rel` variable before transforming it (`rel = rel_for_filtering`). In this method, you transform it directly on the first call, and this line and the next have different argument names, making it in my opinion less clear. Also, the ruby style guide recommends to "avoid shadowing methods with local variables unless they are both equivalent" (we'll have some refactoring to do there).

Personally, I think that if a method is called `apply_some_filter` and takes one argument, it's superfluous to call that argument `something_for_filtering` when you could just call it `something`. It obvious that it's for filtering. 

I think this format would be clearer:

```
def apply_some_filter rel
  filtered_rel = rel

  # Filter by ...
  filtered_rel = rel.where(...)

  # Group by ... if ...
  if some_condition
    filtered_rel = filtered_rel.group(...)
  end

  filtered_rel
end
```

And finally, I would call the two previous methods `apply_some_filter` instead of `apply_some_filters` (remove the "s" at the end). There's multiple users and projects, but each of those methods applies one filter. This of course doesn't apply to this method.

app/api/probe_dock/metrics_api.rb
```diff
@@ -44,13 +44,19 @@ def current_organization
           end
         end
 
-        def apply_projects_and_users_filters(rel_for_filtering)
+        def apply_projects_filters(rel_for_filtering)
           rel = rel_for_filtering
 
           if params[:projectIds].present? && params[:projectIds].kind_of?(Array)
             rel = rel.where 'projects.api_id IN (?)', params[:projectIds].collect(&:to_s)
```
AlphaHydrae: Method calls with arguments should have parentheses (in several places in this file). We should refactor old code as we add new features.

app/api/probe_dock/metrics_api.rb
```diff
@@ -72,6 +84,17 @@ def count_new_tests_by_interval(start_date, interval)
 
           rel.to_a.inject({}){ |memo,data| memo[data.project_tests_by_interval.strftime('%Y-%m-%d')] = data.project_tests_count; memo }
         end
+
+        def project_version
+          # Retrieve the data from the specified version otherwise take the most recent project version to retrieve the data
+          if params[:projectVersionId].present?
+            ProjectVersion.where('api_id = ?', params[:projectVersionId]).first
```
AlphaHydrae: I would use `first!` here. I think we want this call to fail if provided with an unknown ID.

app/api/probe_dock/metrics_api.rb
```diff
@@ -44,13 +44,19 @@ def current_organization
...
           end
 
+          rel
+        end
+
+        def apply_users_filters(rel_for_filtering)
+          rel = rel_for_filtering
+
           if params[:userIds].present? && params[:userIds].kind_of?(Array)
            users = User.where('api_id IN (?)', params[:userIds].collect(&:to_s)).to_a
```
AlphaHydrae: Indentation is wrong in this `if`.

app/api/probe_dock/metrics_api.rb
```diff
@@ -228,6 +245,77 @@ def count_new_tests_by_interval(start_date, interval)
...
+        filter_by_projects = params[:projectIds].present? && params[:projectIds].kind_of?(Array)
+        filter_by_users = params[:userIds].present? && params[:userIds].kind_of?(Array)
+
+        # Try to retrieve the project version
+        project_version = project_version()
+
+        project_versions = if !project_version.nil?
+          [ project_version.id ]
+        else
+          # Prepare a statement (postgresql) to retrieve the latest version by project
```
AlphaHydrae: Be careful with terminology. PostgreSQL has prepared statements (similar to MySQL's as far as I know), but we're not preparing a statement here. We're just dynamically building a query that will be executed later. It is not optimized like a prepared statement would be.

app/assets/stylesheets/modules/tests-by-category-widget.styl
```diff
@@ -0,0 +1,32 @@
+.tests-by-category-table
```
AlphaHydrae: Avoid creating several top-level styles for the same widget:

```
.tests-by-category-table
  ...
.tests-by-category-td
  ...
.tests-by-category-progress
  ...
```

Add a class or ID to the widget, and scope the other classes underneath. Also, prefer semantic descriptions to direct references to HTML tags:

```
.tests-by-category-widget
  .categories-list
    ...
  .description
    ...
  .progress
    ...
```

app/views/templates/tests-by-category-widget.html.slim
```diff
@@ -0,0 +1,66 @@
+.panel.panel-default
+  .panel-heading
+    .row
+      .col-md-6
```
AlphaHydrae: This `.row` and the two `.col-md-6` are completely superfluous. Also, the button drops below the title when the screen is smaller.

Put the right-floating button before the `h3` (otherwise it's displayed below it) and add `.clearfix` to the `.panel-heading`.

app/views/templates/tests-by-category-widget.html.slim
```diff
@@ -0,0 +1,66 @@
...
+                        .btn-group
+                          button.btn.btn-warning ng-click='resetFilters()' Reset
+                        .btn-group
+                          button.btn.btn-primary ng-click='closeFilters()' Close
+
+  .panel-body ng-if='testsByCategory.length === 0'
+    .empty
+      em Loading...
+  .panel-body ng-if='testsByCategory.length > 0'
+    table.table.tests-by-category-table
```
AlphaHydrae: It's a bad idea to use Bootstrap's `.table` here if you're going to override all its styles so that it looks nothing like Bootstrap's table. Simply use a normal table so that we only have to add the custom style we need.

app/views/templates/tests-by-category-widget.html.slim
```diff
@@ -0,0 +1,66 @@
...
+      .col-md-6
+        h3.panel-title Tests by category
+      .col-md-6
+        div style='float:right;'
+          a.btn.btn-default.btn-xs href='#' role='button'
+            i.glyphicon.glyphicon-filter
+            drop constrain-to-scroll-parent='true' constrain-to-window='true' open-on='"click"' position='"left top"' tether-options='{offset: "-5px -5px"}' before-close='beforeClose()'
+              .panel.panel-default.tests-by-category-filters
+                .panel-heading
+                  .h4.panel-title Filters
```
AlphaHydrae: `.h4`

PR #27, opened 2016-02-10
Test results api

Prevole:
Not ready yet!


Prevole:
Added the filter by test on the project versions API. The API tests is quite slow but enforce the scoping of data.

app/api/probe_dock/project_versions_api.rb
```diff
@@ -26,6 +26,8 @@ def current_organization
             Organization.active.where(normalized_name: params[:organizationName].to_s.downcase).first!
           elsif params[:projectId].present?
             Organization.active.joins(:projects).where('projects.api_id = ?', params[:projectId].to_s).first!
+          elsif params[:testId].present?
+            ProjectTest.where(api_id: params[:testId]).first!.project.organization
```
AlphaHydrae: This makes 2 queries and loads data that will not be used. The relation should start from Organization and make the necessary joins like the other queries in this method. It also allows you to use `Organization.active` which makes sure that it only works for active organizations. Please also update `ResultsApi#current_organization` to use this pattern.

app/api/probe_dock/results_api.rb
```diff
@@ -0,0 +1,82 @@
...
+            ProjectTest.where(api_id: params[:testId]).first!.project.organization
+          end
+        end
+      end
+
+      get do
+        authorize! TestResult, :index
+
+        rel = TestResult
+          .order('test_results.active desc, test_results.passed, test_results.name, test_results.id')
```
AlphaHydrae: This order clause was meant for display to display the results in test reports, where we display failing results first, then passing ones, then inactive ones.

In the case of the test widget, we'll want to order by run date.

I would leave this as the default order, but allow the user to give a `sort=runAt` param which will change the order.

AlphaHydrae: This order clause was meant for display to display the results in test reports, where we display failing results first, then passing ones, then inactive ones.

In the case of the test widget, we'll want to order by run date.

I would leave this as the default order, but allow the user to give a `sort=runAt` param which will change the order.

client/pages/report-details/details.reportResults.directive.js
```diff
@@ -81,7 +81,7 @@ angular.module('probedock.reportDetailsPage').directive('reportDetailsResults',
       $scope.fetchingMoreResults = true;
 
       return api({
-        url: '/reports/' + $scope.report.id + '/results',
+        url: '/results?reportId=' + $scope.report.id,
```
AlphaHydrae: Please pass the `reportId` query parameter in the `params` object.

spec/support/named_records_spec_helper.rb
```diff
@@ -16,15 +16,55 @@
...
+  end
+
+  def named_record_by_type(name, type)
+    key = NamedRecordKey.new(name, type)
+    record = @named_records.try(:[], key)
+    raise "Unknown named record #{key}" unless record
     record
   end
+
+  class NamedRecordKey
```
AlphaHydrae: Remove if unused.

app/api/probe_dock/project_versions_api.rb
Prevole: Done

app/api/probe_dock/results_api.rb
Prevole: Done

client/pages/report-details/details.reportResults.directive.js
Prevole: Done

spec/support/named_records_spec_helper.rb
Prevole: Done (kept in GIT history :))

PR #29, opened 2016-02-13
Test results widget part 2

Prevole:
Not yet ready.


app/api/probe_dock/metrics_api.rb
```diff
@@ -443,6 +443,20 @@ def current_project_version
           contribution
         end
       end
+
+      get :versionsWithNoResult do
+        authorize!(:organization, :data)
+
+        project_versions_rel = policy_scope(ProjectVersion).order('project_versions.created_at DESC')
+        project_versions_rel = project_versions_rel.joins(project: [:tests, :organization])
```
AlphaHydrae: I would avoid reassignments if there's no reason to do it and no useful comments in between. Simply write:

```
project_versions_rel = policy_scope(ProjectVersion)
  .order(...)
  .joins(...)
  .where(...)
```

app/api/probe_dock/project_versions_api.rb
```diff
@@ -50,7 +50,7 @@ def current_organization
           end
 
           if params[:testId].present?
-            paginated_rel = paginated_rel.joins(test_results: :test).where('project_tests.api_id = ?', params[:testId])
+            paginated_rel = paginated_rel.joins(test_results: :test).where('project_tests.api_id = ?', params[:testId]).group('project_versions.id, projects.name')
```
AlphaHydrae: Usually when there's a relation that has to have a group clause depending on the parameters, I prefer to use a boolean `group` variable like this:

```
paginated rel do |rel|
  group = false

  if params[:a]
    group = true
    ...
  end

  if params[:b]
    ...
  end

  if params[:c]
    group = true
    ...
  end

  rel = rel.group(...) if group
end

It avoids repeating the group clause if multiple many-to-many joins make it necessary (since it's usually always the same group clause). Even if there's only one parameter that requires it at the moment, doing it like this discourages code duplication.

Comment (AlphaHydrae) [inline app/api/probe_dock/project_versions_api.rb]: I didn't see it before, but avoid joining through `:test_results` at all costs. It will always be slow, and we might remove that table in the future (if we move the data to another DB like elastic search). You can obtain the same result by joining through `:test_descriptions`.

File: client/components/project-version-select/select.template.html.slim [hunk 3]
```diff
@@ -9,7 +9,7 @@
 
   / dropdown menu to select an existing version
   ui-select id='{{ prefix }}-project-version-select' ng-if='!config.newVersion && modelProperty' ng-model='modelObject[modelProperty]' reset-search-input=true
-    ui-select-match placeholder='{{ getPlaceholder() }}' allow-clear='{{ !!latestVersion }}' {{ $select.selected.name }}
+    ui-select-match placeholder='{{ getPlaceholder() }}' allow-clear='{{ allowClear || !!latestVersion }}' {{ $select.selected.name }}
```
AlphaHydrae: If we now have an explicit `allowClear` option for the directive, then I think that if it's set, it alone should determine whether the clear button is shown. That would be more intuitive.

You could set a default depending on whether the latest version is given, but if I use this directive with `allow-clear='false'`, I expect no clear button to be shown.

client/components/result-status-icon/icon.directive.js
```diff
@@ -0,0 +1,21 @@
...
+    scope: {
+      result: '='
+    }
+  };
+}).controller('ResultStatusIconCtrl', function($scope) {
+  $scope.ready = false;
+
+  $scope.$watch('result', function (value) {
+    if (value) {
+      $scope.passed = ($scope.result.passed || $scope.result.passing) && $scope.result.active;
```
AlphaHydrae: If this directive can display the result for a test or one of the results of that test, then it should be named `test-status-icon` and have two possible arguments: either a test or a result. Passing a test in an argument named `result` is quite counter-intuitive.

Or if it's only used for results, then you can remove the `passing` fallback.

client/components/result-status-icon/icon.directive.js
```diff
@@ -0,0 +1,21 @@
...
+      $scope.failed = !($scope.result.passed || $scope.result.passing) && $scope.result.active;
+      $scope.active = !$scope.result.active;
+      $scope.ready = true;
+    }
```
AlphaHydrae: If the directive is used only for results, you can actually remove this whole `$watch`. It would be unnecessary as you can simply use `result.passed` and `result.active` in the template.

client/components/user-select/select.directive.js
```diff
@@ -7,7 +7,9 @@ angular.module('probedock.userSelect').directive('userSelect', function() {
       organization: '=',
       modelObject: '=',
       modelProperty: '@',
-      prefix: '@'
+      prefix: '@',
+      filterByLabel: '@',
+      filterAllLabel: '@'
```
AlphaHydrae: These scope argument names don't sound very intuitive to me considering how they're used. I would prefer `label` and `placeholder`, or `fieldLabel` and `fieldPlaceholder`, for example. It better expresses their function.

client/initializers/underscore.js
```diff
@@ -15,3 +15,69 @@ _.isPresent = function(value) {
 _.isBlank = function(value) {
   return !_.isPresent(value);
 };
+
+// FIXME: Replace this comparator function by server side implementation
+// Based on topic: http://stackoverflow.com/questions/6832596/how-to-compare-software-version-number-using-js-only-number
+_.compareVersion = function(v1, v2) {
```
AlphaHydrae: I don't really care, but I would have expected these functions to be defined in a `projectVersions` angular service.

client/initializers/underscore.js
```diff
@@ -15,3 +15,69 @@ _.isPresent = function(value) {
...
+  var versionRegex = /(\d+)?(.*)?/;
```
AlphaHydrae: This is not a very clean way to match a semantic versioning component. It would match an empty string, or a string with both a number and text like "42foo", neither of which are valid. I would first check whether each part is a number with a simple `\d+` regexp, and then run the appropriate comparison depending on whether it's a number or not.

There are a lot of other ways in which this function would not sort semantic versions correctly, but we can leave that for another time. I have added a ticket to our board for the server-side implementation.

client/widgets/test-results/results.directive.js
```diff
@@ -0,0 +1,118 @@
...
+
+      if (!$scope.missingVersions) {
+        return api({
+          url: '/metrics/versionsWithNoResult',
+          params: {
+            organizationId: $scope.organization.id,
+            testId: $scope.test.id
+          }
+        }).then(function(response) {
+          if (response.data) {
```
AlphaHydrae: This check makes little sense. It will always pass with the current implementation. I would remove it.

client/widgets/test-results/results.directive.js
```diff
@@ -0,0 +1,118 @@
...
+    }).then(function(response) {
+      if (response.data) {
+        $scope.pagination = response.pagination();
+
+        // Reverse and prepend the test results
+        $scope.results = response.data.reverse().concat($scope.results);
+
+        // Calculate the next number of results to load
+        var remainingResults = $scope.pagination.total - $scope.results.length;
+        $scope.nextChunk = remainingResults > pageSize ? pageSize : remainingResults;
```
AlphaHydrae: You already store the pagination data in the scope. Is there really a need to redo a calculation to check whether there are more results to load?

client/widgets/test-results/results.template.html.slim
```diff
@@ -0,0 +1,30 @@
+.test-results-widget.panel.panel-default
+  .panel-heading
+    help-button
+      p Shows results for different versions. By default, the results for all the versions are displayed.
+      p The list of versions shows only the versions for which results are available. The versions without results are not displayed.
```
AlphaHydrae:
```
This widget shows the test results received for this specific test in chronological order (from older to newer). By default, only the latest 50 results are shown. If you choose to load older results, they will be added to the beginning of the list as they are older.

By default, results are shown for all versions of the project. You may use the version dropdown to see only the test results received for a specific version. If there are versions for which the test was never run, they will not be available to filter, but will be listed below the results.
```

client/widgets/test-results/results.directive.js
Prevole: Wrong! Avoid calling the API each time a new chunk of results is loaded.

client/widgets/test-results/results.directive.js
Prevole: I think so, I want to know how many records will be loaded next time.

client/widgets/test-results/results.directive.js
AlphaHydrae: This is performed by the check on line 103.

If `response.data` is an array, empty or not, this check will always pass.

client/widgets/test-results/results.template.html.slim
```diff
@@ -0,0 +1,30 @@
...
+  .panel-body
+    p.empty ng-if='!loading && results.length === 0' Test has no result.
+    .row ng-if='loading || results.length > 0'
+      .col-md-12
+        p ng-if='loading' Loading...
+        p ng-if='!loading'
+          a.pull-right.more href='#' ng-if='!loading && nextChunk > 0' ng-click='fetchMore()'
+            ng-pluralize count='nextChunk' when='{"1": "Load one more result...", "other": "Load {} more results..."}'
+          span ng-if='results.length == 1' There is only one available result.
+          span ng-if='results.length > 1 && results.length < pagination.total' The {{ results.length }} recent results on the {{ pagination.total }} available results.
```
AlphaHydrae: ```Showing the latest X results ouf of Y available results.```

client/widgets/test-results/results.template.html.slim
```diff
@@ -0,0 +1,30 @@
...
+          span ng-if='results.length == pagination.total' All the {{ results.length }} available results.
```
AlphaHydrae: ```Showing all X results.```

client/widgets/test-results/results.directive.js
AlphaHydrae: Agreed. Comment due to sequential review of files.

client/widgets/test-results/results.directive.js
Prevole: Yeps, agree. Was thinking of ` if (!$scope.missingVersions)`. Don't ask why :)

features/metrics_versions_with_no_result.feature
```diff
@@ -0,0 +1,112 @@
+@api @metrics
+Feature: Versions with no result for a test
+
+  Retrieve the versions where there is no result for the test.
```
AlphaHydrae: If your description of the feature is going to be the same as the Cucumber feature title, don't write one. In this case, I would explain why it's useful to retrieve the versions where there are no results: it may allow you to identify tests that may not be running due to a technical issue.

client/widgets/test-results/results.directive.js
AlphaHydrae: While we're at it, it would be cleaner to move this call out of `fetchResults` and put it in a `fetchMissingVersions`, which would negate the need for this check and run both API calls in parallel.

PR #36, opened 2016-02-18
Test execution time widget

Prevole:


client/widgets/test-execution-time/time.directive.js
```diff
@@ -0,0 +1,124 @@
...
+  });
+
+  $scope.$watch('params', function(value) {
+    if (!_.isEmpty(value)) {
+      initFetchResults();
+    }
+  }, true);
+
+
+  $scope.fetchMore = function() {
```
AlphaHydrae: As I said before, I don't think it's a good idea to allow to fetch more results for this widget. A chart has a very limited number of results it can display before they become too close together and unreadable, unlike the test results widget were they wrap. For now, I would display a fixed number of latest results, and later add functionality to allow to change the time window (or use a zoomable time series). But a simple chart like this doesn't lend itself to displaying all results.

AlphaHydrae: Addendum: I hadn't seen there's a maximum. However, is 100 results in the chart readable on anything other than a full screen with 1920+ pixels resolution? We can leave it in for now, but in general I'm not convinced it's a good idea to implement such unresponsive display mechanisms in small widgets. I'd rather save the time and use it to implement a better solution later.

client/widgets/test-execution-time/time.directive.js
```diff
@@ -0,0 +1,124 @@
...
+      params.runnerIds = $scope.params.userIds;
+    }
+
+    $scope.loading = true;
+
+    return api({
+      url: '/results',
+      params: params
+    }).then(function(response) {
+      if (response.data) {
```
AlphaHydrae: Useless check. Will always be true.

client/widgets/test-execution-time/time.directive.js
```diff
@@ -0,0 +1,124 @@
...
+  $scope.params = {};
+  $scope.results = [];
+  $scope.loading = true;
+
+  $scope.chart = {
+    data: [],
+    labels: [],
+    options: {
+      pointHitDetectionRadius: 5,
+      showTooltips: false,
```
AlphaHydrae: Why? I would expect to be able to hover the mouse over the chart points to see the date, time and exact execution time.

client/widgets/test-execution-time/time.template.html.slim
```diff
@@ -0,0 +1,30 @@
...
+        canvas.chart.chart-line height='150' ng-if='chart.data.length' data='chart.data' labels='chart.labels' options='chart.options'
+        .pull-left
+          .text-center {{ results[0].runAt | amDateFormat: 'HH:mm:ss' }}
+          .text-center {{ results[0].runAt | amDateFormat: 'DD.MM.YYYY' }}
+        .pull-right
+          .text-center {{ results[results.length - 1].runAt | amDateFormat: 'HH:mm:ss' }}
+          .text-center {{ results[results.length - 1].runAt | amDateFormat: 'DD.MM.YYYY' }}
+    .row
+      .col-md-6
+        project-version-select prefix='test-results' organization='organization' test='test' model-object='params' allow-clear='true' placeholder='All versions'
```
AlphaHydrae: Unrelated to this feature, but I noticed that your semantic versioning sort has changed the order in which versions are displayed by this directive. They used to be displayed from most recent to oldest, now it's the other way around. I think the most recent versions are going to be the most useful, especially when there's a lot. It should be fixed.

client/widgets/test-execution-time/time.directive.js
AlphaHydrae: Addendum 2: tested locally. It seems to look good, at least without the dots that other charts have, but I think I would actually like it better if it simply displayed 100 results outright. It seems to me that the most results are displayed, the better chance you have of seeing a trend in the execution time, which is the purpose of this widget. Showing only 25 by default seems like a waste.

client/widgets/test-execution-time/time.directive.js
```diff
@@ -0,0 +1,124 @@
...
+      datasetFill: false,
+      scaleShowVerticalLines: false,
+
+      /*
+       * Fix for space issue in the Y axis labels
+       * see: https://github.com/nnnick/Chart.js/issues/729
+       * see: http://stackoverflow.com/questions/26498171/how-do-i-prevent-the-scale-labels-from-being-cut-off-in-chartjs
+       */
+      scaleLabel: function (object) {
+        return '  ' + object.value;
```
AlphaHydrae: See if you can use [this filter](https://github.com/probedock/probedock/blob/master/client/filters/formatDuration.filter.js) to format the value. It's quite configurable (although it lacks documentation). The `maxParts` argument should avoid an overly long string if set to 1. You shouldn't need the last two.

client/widgets/test-execution-time/time.template.html.slim
```diff
@@ -0,0 +1,30 @@
+.test-execution-time-widget.panel.panel-default
+  .panel-heading
+    help-button
+      p This widget shows the execution time for this specific test in chronological order (from older to newer). By default, only the latest {{ pageSize }} results are shown. If you choose to load older results, they will be added to the beginning of the curve as they are older. Only a maximum of {{ maxResults }} results can be shown.
```
AlphaHydrae: Replace first two sentences with "This widget shows the execution time history for this specific test in chronological order (from older runs to newer). By default, it only shows the execution time of the latest {{ pageSize }} results."

client/widgets/test-execution-time/time.template.html.slim
```diff
@@ -0,0 +1,30 @@
...
+      p By default, results are shown for all versions of the project. You may use the version dropdown to see only the test results received for a specific version. If there are versions for which the test was never run, they will not be available to filter.
```
AlphaHydrae: "By default, the execution time history is shown across all versions of the project. You may use the version dropdown to display only the history of a specific version. If there are versions for which the test was never run, they will not be available to filter."

client/widgets/test-execution-time/time.template.html.slim
```diff
@@ -0,0 +1,30 @@
...
+    .row.chart-row ng-if='loading || results.length > 0'
+      .col-md-12
+        p ng-if='loading' Loading...
+        p ng-if='!loading'
+          a.pull-right.more href='#' ng-if='!loading && nextChunk > 0' ng-click='fetchMore()'
+            ng-pluralize count='nextChunk' when='{"1": "Load one more result...", "other": "Load {} more results..."}'
+          span ng-if='results.length == 1' There is only one available result.
+          span ng-if='results.length > 1 && results.length < total' Showing the latest {{ results.length }} results ouf of a maximum of {{ total }} results.
+          span ng-if='results.length == total && total == pagination.total' Showing all {{ results.length }} results.
+          span ng-if='results.length == total && total < pagination.total' Reached the maximum of {{ results.length }} results.
```
AlphaHydrae: Again, personally I think this widget would look better if it simply displayed 100 results and said "Showing the execution time of the latest X test results."

Whether you change this or not, it might be a good idea not to show the chart if there is only 1 result, and instead to display a message like "This test was only run once and took Xms to run." (Use the formatDuration filter.)

AlphaHydrae:
It could be interesting to add text indicating the average value (and maybe also the min/max which might not be obvious to see if there's a lot of data in the chart).

PR #39, opened 2016-02-22
Project repo link

Prevole:


app/models/project.rb
```diff
@@ -31,6 +31,7 @@ class Project < ActiveRecord::Base
   validates :name, presence: true, uniqueness: { scope: :organization_id, case_sensitive: false }, length: { maximum: 50, allow_blank: true }, format: { with: /\A[a-z0-9]+(?:\-[a-z0-9]+)*\Z/i }
   validates :display_name, length: { maximum: 50, allow_blank: true }
   validates :organization, presence: true
+  validates :repo_url, url: { allow_blank: true }, length: { maximum: 100, allow_blank: true }
```
AlphaHydrae: A limit of 100 seems potentially short for a URL of arbitrary length. We have URLs that are ~80 characters in the SoftEng organization, so it's not a stretch to imagine people having longer URLs. What do you think?

AlphaHydrae:
I did not like that we had an additional `?id` query parameter in the new `org.projects.show.edit` route. when the route already contains everything we need to find the project, and it's already loaded by the controller of the page.

I refactored the edit project modal service a little to allow passing the project or project ID as an option. I also moved the state change callback to be registered only after the project has been loaded (using our existing `states` service).

See https://github.com/probedock/probedock/commit/cfb14910e5cc19cf65124aa35b045c065c2f329d

client/pages/project-details/details.template.html.slim
```diff
@@ -25,6 +30,9 @@
               dd {{ project.createdAt | amDateFormat:'llll' }}
               dt Last Update
               dd {{ project.updatedAt | amDateFormat:'llll' }}
+              dt ng-if='project.repoUrl' Code repository
+              dd ng-if='project.repoUrl'
+                a href='{{ project.repoUrl }}' target='_blank' {{ project.repoUrl }}
```
AlphaHydrae: This doesn't look good when the URL is too long, since URLs don't wrap. We should add some CSS to truncate if it overflows.

client/pages/project-details/details.template.html.slim
```diff
@@ -2,8 +2,13 @@
   .page-header
     h1
       em Loading project data...
-.container-fluid ng-if='project && currentOrganization'
+.container-fluid.project-details ng-if='project && currentOrganization'
   .page-header
+    .btn-group.pull-right role='group' aria-label='Project Controls'
+      a.btn.btn-primary ng-if='project.repoUrl' href='{{ project.repoUrl }}' tooltip='Open repository' tooltip-append-to-body='true' target='_blank'
+        project-repo-icon project='project'
+      button.btn.btn-warning tooltip='Edit' tooltip-append-to-body='true' ui-sref='org.projects.show.edit({ projectName: project.name, id: project.id })' ng-class='{active: currentState == "org.projects.show.edit"}'
```
AlphaHydrae: You can achieve the same as `ng-class='{active: currentState == "org.projects.show.edit"}'` on this button with `ui-sref-active=true`.

client/pages/project-edit-modal/modal.template.html.slim
```diff
@@ -17,6 +17,11 @@ form novalidate=true role='form' name='projectForm' ng-submit='projectForm.$vali
       p.help-block
         | The is the name of the project as shown in ProbeDock.
     .form-group
+      label for='projectRepoUrl' Repository URL
+      input#projectRepoUrl.form-control type='string' placeholder='{{ editedProject.repoUrl || "Project repository" }}' maxlength=100 ng-model='editedProject.repoUrl'
```
AlphaHydrae: Why is `editedProject.repoUrl` in the placeholder? As far as I can see it doesn't do anything, as when the placeholder is shown, it means that `editedProject.repoUrl` is empty anyway.

AlphaHydrae: If what you wanted to do was show the original URL in the placeholder when the field is empty, I think that's a bad idea. That would make you think that the URL would remain if you pressed the submit button, when it will actually be removed.

AlphaHydrae:
We could add the link to the project list as well.

app/models/project.rb
Prevole: Agreed!

client/pages/project-details/details.template.html.slim
Prevole: Done.

client/pages/project-details/details.template.html.slim
Prevole: Great! Was not aware.

client/pages/project-edit-modal/modal.template.html.slim
Prevole: Fixed

PR #41, opened 2016-02-26
Technical user in widgets

Prevole:
Cannot be merged before PR <= 36 is on master and FIXME are fixed in this PR.


client/widgets/project-health/health.directive.js
```diff
@@ -1,68 +1,16 @@
...
   return {
     restrict: 'E',
-    controller: 'ProjectHealthWidgetCtrl',
     templateUrl: '/templates/widgets/project-health/health.template.html',
+    controller: 'ProjectHealthWidgetCtrl',
     scope: {
-      project: '='
+      organization: '=',
+      project: '=',
+      linkToVersion: '='
```
AlphaHydrae: Why is this called `linkToVersion` and not `linkable` like the others? What does it mean? It's not clear at all.

client/widgets/test-run-state/state.directive.js
```diff
@@ -0,0 +1,184 @@
+angular.module('probedock.testRunStateWidget').directive('testRunStateWidget', function() {
```
AlphaHydrae: I don't like the name very much. It sounds like this is a widget to show the state of one test run, but it does much more than that since it displays the project health. It's the same code that you extracted from the project health widget.

I would suggest two alternatives:

1) Put the code back in the project health widget and add a scope argument that can be used to switch between the full template with a Bootstrap panel, or the simplified template (the full template should be able to include the simplified one with an `ng-include`, avoiding duplication).

2) Rename this widget `compactProjectHealthWidget` or another similar prefix like "simple" or "bare".

What do you think?

Prevole: Applied the first solution, it is probably a better pattern for next situations like this.

client/widgets/project-health/health.directive.js
Prevole: Done

client/widgets/project-health/health.directive.js
```diff
@@ -1,68 +1,186 @@
...
+   */
+  function fetchReport() {
+    $scope.loading = true;
+
+    // Use the reports API if there is any criteria
+    var url, params;
+    if ($scope.params.projectVersionId || $scope.params.runnerId) {
+      url = '/reports';
+      params = buildReportsParams();
+    } else if ($scope.project.lastReportId) { // Lookup by lastReportId
```
AlphaHydrae: This looks weird to me. Couldn't you simply update `buildReportParams` to always send the project ID, thus avoiding the need for this alternative request?

As a bonus, you would also no longer need `_.isArray(res.data) ? res.data[0] : res.data`

client/widgets/project-health/health.directive.js
AlphaHydrae: Same in the recent activity directive.

PR #45, opened 2016-03-11
Refactoring of the select components to share as much as possible of …

Prevole:
…the same code and attributes


Prevole:
Ready for review

client/components/select/select.directive.js
```diff
@@ -0,0 +1,240 @@
...
+ * @param $scope The scope
+ * @param api The API service
+ * @param options The options for the controller configuration
+ */
+function setupController($scope, api, options) {
+  if (!$scope.prefix) {
+    throw new Error("The prefix attribute on select directive is not set.");
+  }
+
+  if (!options) {
```
AlphaHydrae: Personally when I have sequences of validations like this, I use `else if` and don't put new lines in between. It's not interesting or complicated code so I don't want it taking space when the implementation below is more interesting.

client/components/select/select.directive.js
```diff
@@ -0,0 +1,240 @@
...
+    throw new Error("No options defined to setup the select controller.");
+  }
+
+  if (!options.defaults) {
+    throw new Error("No defaults provided to setup the select controller.");
+  }
+
+  if (!options.fetchUrl) {
+    throw new Error("The fetchUrl must be provided.");
```
AlphaHydrae: When you add a 5th check for another option in the future, do you think you can use yet another phraseology to describe the error? You've managed to write 4 error messages for missing options in 4 different ways.

client/components/select/select.directive.js
```diff
@@ -0,0 +1,240 @@
...
+.directive('userSelect', function() { return createSelectDirective('UserSelectCtrl', defaultsAttributes); })
+
+// Select controllers
+.controller('CategorySelectCtrl', function(api, $scope) {
+  setupController($scope, api, {
+    defaults: {
+      modelProperty: $scope.multiple ? 'categoryName' : 'categoryNames',
+      label: 'Filter by category',
+      placeholder: 'All categories',
+      labelNew: 'create new category.',
```
AlphaHydrae: Is there a reason you added a dot at the end of the label?

client/components/select/select.directive.js
```diff
@@ -0,0 +1,240 @@
...
+  });
+
+  // Define a function to show the item value in the select. If no function
+  // is defined, a lookup on a property given by showProperty is done.
+  if (options.showItem) {
+    $scope.showItem = options.showItem;
+  } else {
+    $scope.showItem = function(item) {
+      return item[$scope.showProperty];
+    }
```
AlphaHydrae: Missing semicolon.

client/components/select/select.directive.js
```diff
@@ -0,0 +1,240 @@
...
+      $scope.modelObject[$scope.modelProperty] = previousItem ? previousItem : $scope.defaultNewItem;
+    } else if (value === false && $scope.choices.length && $scope.autoSelect) {
+      // auto-select the first existing item when disabling creation of a new item
+      $scope.modelObject[$scope.modelProperty] = $scope.choices[0][$scope.extract];
+    }
+  });
+
+  // Define a function to show the item value in the select. If no function
+  // is defined, a lookup on a property given by showProperty is done.
+  if (options.showItem) {
```
AlphaHydrae: I don't like the `show` naming very much. In angular it's a term often use in relation with visibility. I would advise renaming this to `formatItem`.

AlphaHydrae: In that case, `showProperty` could be renamed to `displayedProperty`.

client/widgets/test-key-generator/generator.template.html.slim
```diff
@@ -16,14 +16,15 @@
     form.form-inline name='generatorForm' novalidate=true ng-submit='generate()'
       button.btn.btn-success type='submit' ng-disabled='generatorForm.$invalid'
         span.glyphicon.glyphicon-plus
-        |  Generate 
+        |  Generate
```
AlphaHydrae: Please add the relevant `>` and `<` to replace the trailing spaces your editor deleted.

PR #50, opened 2016-03-22
Management page (widgets)

Prevole:


app/api/probe_dock/platform_management_api.rb
```diff
@@ -0,0 +1,51 @@
...
+          @current_organization ||= if params[:organizationId].present?
+            Organization.active.where(api_id: params[:organizationId].to_s).first!
+          elsif params[:organizationName].present?
+            Organization.active.where(normalized_name: params[:organizationName].to_s.downcase).first!
+          end
+        end
+      end
+
+      get :dbStats do
+        authorize!(:platform_table_info, :show)
```
AlphaHydrae: Authorization calls for collections should have the action `:index` instead of `:show`.

app/models/organization_table_info.rb
```diff
@@ -0,0 +1,165 @@
...
+    # Retrieve the results count, payloads count and organization info for the top most consuming organizations
+    org_payloads_and_results_count = Organization
+      .select('
+        organizations.id, organizations.api_id, organizations.name, organizations.display_name,
+        SUM(test_payloads.results_count) AS test_results_count, COUNT(test_payloads.id) as test_payloads_count
+      ')
+      .joins('
+        LEFT OUTER JOIN "projects" ON "projects"."organization_id" = "organizations"."id"
+        LEFT OUTER JOIN "project_versions" ON "project_versions"."project_id" = "projects"."id"
+        LEFT OUTER JOIN"test_payloads" ON "test_payloads"."project_version_id" = "project_versions"."id"
```
AlphaHydrae: Missing space: `JOIN"test_payloads"`.

app/models/organization_table_info.rb
```diff
@@ -0,0 +1,165 @@
...
+      ')
+      .group('organizations.id')
+      .order('test_results_count DESC NULLS LAST')
```
AlphaHydrae: For the rare but possible cases where organizations have the same results count (or no results), I would add a second order criteria on the name to have predictable ordering.

app/models/organization_table_info.rb
```diff
@@ -0,0 +1,165 @@
...
+    org_results_trends_count = Organization
+      .select("
+        organizations.id,
+        SUM(test_payloads.results_count) AS test_results_count,
+        date_trunc('week', test_payloads.processed_at) AS trend_date
+      ")
+      .joins('
+        LEFT OUTER JOIN "projects" ON "projects"."organization_id" = "organizations"."id"
+        LEFT OUTER JOIN "project_versions" ON "project_versions"."project_id" = "projects"."id"
+        LEFT OUTER JOIN"test_payloads" ON "test_payloads"."project_version_id" = "project_versions"."id"
```
AlphaHydrae: Same missing space.

app/models/organization_table_info.rb
```diff
@@ -0,0 +1,165 @@
+class OrganizationTableInfo
+  attr_accessor :organizations_counts, :total_counts
+
+  def initialize(org_stats, total_counts)
+    @organizations_counts = org_stats.kind_of?(Array) ? org_stats : [ org_stats ]
+    @total_counts = total_counts
+  end
+
+  def self.top_stats(top: 0, organization: nil, weeks: 5)
```
AlphaHydrae: Maybe `weeks` should be renamed to `trends_weeks` for clarity.

app/models/organization_table_info.rb
```diff
@@ -0,0 +1,165 @@
...
+      .where("date_trunc('week', test_payloads.processed_at) >= ?", weeks.weeks.ago.at_beginning_of_week)
+
+    # Add where clause to scope to specific organization
+    if organization
+      org_payloads_and_results_count = org_payloads_and_results_count.where('organizations.id = ?', organization.id)
+      org_results_trends_count = org_results_trends_count.where('organizations.id = ?', organization.id)
+    end
+
+    # Add a limit to do the top n
+    if top > 0
```
AlphaHydrae: I don't like having `top` default to 0 and list all organization stats. It's like having a paginated resource that returns all results with no pagination if you don't pass parameters. I would set the default to 5 like we're using now, and I would not allow passing 0 to return all the stats (just like we do with pagination).

app/models/organization_table_info.rb
```diff
@@ -0,0 +1,165 @@
...
+        # Check if the DB returned data for the week date, otherwise consider as 0
+        results_trend[org_id] << (week_trend.nil? ? 0 : week_trend.test_results_count)
+      end
+    end
+
+    # Retrieve the count data for projects
+    organizations_projects_count = Organization
+      .select('organizations.id as org_id, COUNT(projects.id) as projects_count')
+      .joins('LEFT OUTER JOIN "projects" ON "projects"."organization_id" = "organizations"."id"')
+      .where('organizations.id in (?)', org_ids)
```
AlphaHydrae: Following the convention we have applied so far for SQL operations, you should write `IN` in uppercase instead of `in`.

app/models/organization_table_info.rb
```diff
@@ -0,0 +1,165 @@
...
+    @organizations_counts = org_stats.kind_of?(Array) ? org_stats : [ org_stats ]
+    @total_counts = total_counts
+  end
+
+  def self.top_stats(top: 0, organization: nil, weeks: 5)
+    # Retrieve the results count, payloads count and organization info for the top most consuming organizations
+    org_payloads_and_results_count = Organization
+      .select('
+        organizations.id, organizations.api_id, organizations.name, organizations.display_name,
+        SUM(test_payloads.results_count) AS test_results_count, COUNT(test_payloads.id) as test_payloads_count
```
AlphaHydrae: You have a mix of `as` and `AS` in this file's queries. I would suggest always using `AS`.

app/models/organization_table_info.rb
```diff
@@ -0,0 +1,165 @@
...
+        # Try to find the data for the current week date
+        week_trend = org_results_trends.find{ |org_trend| org_trend.trend_date == week_date }
+
+        # Check if the DB returned data for the week date, otherwise consider as 0
+        results_trend[org_id] << (week_trend.nil? ? 0 : week_trend.test_results_count)
+      end
+    end
+
+    # Retrieve the count data for projects
+    organizations_projects_count = Organization
```
AlphaHydrae: Couldn't these project counts and the test counts below be retrieved along with the first query? It already has a join on the `projects` table, it would only need an additional join on `project_tests`.

AlphaHydrae: Addendum: The `organizations` table already has a `projects_count` cache column, and the `projects` table has a `tests_count` cache column which you can sum, so you wouldn't even need additional joins.

app/models/organization_table_info.rb
Prevole: Done with test coverage.

app/models/organization_table_info.rb
```diff
@@ -0,0 +1,165 @@
...
+      org_results_trends_count = org_results_trends_count.where('organizations.id = ?', organization.id)
+    end
+
+    # Add a limit to do the top n
+    if top > 0
+      org_payloads_and_results_count = org_payloads_and_results_count.limit(top)
+    end
+
+    # Collect the ids
+    org_ids = org_payloads_and_results_count.collect(&:id)
```
AlphaHydrae: `org_payloads_and_results_count` is an Active Record relation, not an array of results. Up until this line, the query described by the relation has not yet been executed. When you call `collect` on it, it executes the query and proxies the `collect` method to the result array. Later, at line 95, you again call `each` on the relation, which it also proxies to the result array. It doesn't execute the query twice because the relation hasn't changed so it probably caches the results.

It works but I personally prefer to explicitly transform the relation into a result array:

```
org_payloads_and_results_count_rel = Organization...
org_payloads_and_results_count = org_payloads_and_results_count_rel.to_a
org_ids = org_payloads_and_results_count.collect(&:id)
org_payloads_and_results_count.each do |payloads_and_results_count|
  ...
end
```

It makes it clearer what is an Active Record relation and what is an actual result.

A similar refactoring could be done for the `org_results_trends_count` relation:

```
org_ids.each do |org_id|
  org_results_trends = org_results_trends_count.select{ |org_results_trend| org_results_trend.id == org_id }
  ...
end
```

`select` is called multiple times on the relation, presumably returning the same cached results. But it would be clearer to explicitly retrieve those results before the parent `each`.

app/models/platform_table_info.rb
```diff
@@ -0,0 +1,119 @@
...
+          .joins('
+            LEFT OUTER JOIN "projects" ON "projects"."organization_id" = "organizations"."id"
+            LEFT OUTER JOIN "project_versions" ON "project_versions"."project_id" = "projects"."id"
+            LEFT OUTER JOIN"test_payloads" ON "test_payloads"."project_version_id" = "project_versions"."id"
+          ')
+          .group("date_trunc('week', test_payloads.processed_at)")
+          .order('trend_date ASC')
+          .where("date_trunc('week', test_payloads.processed_at) >= ?", starting_date)
+      else
+        # We prepare the SQL statement
```
AlphaHydrae: Careful with terminology. This is not a prepared SQL statement. It's an SQL query.

app/models/platform_table_info.rb
```diff
@@ -0,0 +1,119 @@
...
+      rows_count_queries[row['name']] = if row['name'] == 'test_results'
+        Organization
+          .select("
+            SUM(test_payloads.results_count) AS count,
+            date_trunc('week', test_payloads.processed_at) AS trend_date
+          ")
+          .joins('
+            LEFT OUTER JOIN "projects" ON "projects"."organization_id" = "organizations"."id"
+            LEFT OUTER JOIN "project_versions" ON "project_versions"."project_id" = "projects"."id"
+            LEFT OUTER JOIN"test_payloads" ON "test_payloads"."project_version_id" = "project_versions"."id"
```
AlphaHydrae: Our friend the missing space is back.

app/models/platform_table_info.rb
```diff
@@ -0,0 +1,119 @@
...
+      INNER JOIN
+        pg_attribute AS a on a.attrelid = c.oid
+      WHERE
+        a.attname = 'created_at'
+        AND c.relkind = 'r'
+    }
+
+    starting_date = weeks.weeks.ago.at_beginning_of_week
+
+    # We want to prepare the SQL queries to count the rows in the retrieved tables
```
AlphaHydrae: I would avoid the term `prepare` in general when describing SQL queries. It's confusing since it has a specific meaning in database terminology. Prefer `build`.

app/models/platform_table_info.rb
```diff
@@ -0,0 +1,119 @@
...
+      FROM (
+        SELECT
+          table_name AS short_table_name,
+          ('"' || table_schema || '"."' || table_name || '"') AS table_name
+        FROM
+          information_schema.tables
+        WHERE
+          table_schema LIKE 'public'
+      ) AS all_tables
+      ORDER BY total_size DESC
```
AlphaHydrae: Add a second order criteria on the name to have predictable ordering of tables with the same total size.

app/models/organization_table_info.rb
```diff
@@ -0,0 +1,165 @@
...
+
+    # Create the array of each organization stats with organization details
+    stats = []
+    org_payloads_and_results_count.each do |payloads_and_results_count|
+      # Retrieve the org_idx corresponding to the idx because we do not want to change the order of org
+      org_idx = organizations_projects_count.index{ |organization_projects_count| organization_projects_count.org_id == payloads_and_results_count.id }
+
+      # Create the stat hash
+      stats << {
+        id: payloads_and_results_count.api_id,
```
AlphaHydrae: I would name this property `api_id` to avoid confusion.

app/api/probe_dock/organizations_api.rb
```diff
@@ -44,50 +44,60 @@ def serialization_options orgs
...
+              rel = rel.joins('LEFT OUTER JOIN memberships ON organizations.id = memberships.organization_id')
+                .where('organizations.public_access = ? OR (memberships.id IS NOT NULL AND memberships.user_id = ?)', true, current_user.id).distinct
             else
-              rel = rel.where 'organizations.public_access = ?', true
+              rel = rel.where('organizations.public_access = ?', true)
             end
           end
 
+          # Retrieve the organization for which the user is an admin
+          if params.key?(:administered)
```
AlphaHydrae: Add `&& !current_user.try(:admin?)`, the same as for the `:accessible` param.

client/widgets/db-stats/stats.directive.js
```diff
@@ -0,0 +1,80 @@
+angular.module('probedock.dbStatsWidget').directive('dbStatsWidget', function() {
+  return {
+    restrict: 'E',
+    controller: 'DbStatsWidgetCtrl',
+    templateUrl: '/templates/widgets/db-stats/stats.template.html',
+    scope: {
+      organization: '='
```
AlphaHydrae: Superfluous parameter.

client/widgets/db-stats/stats.directive.js
```diff
@@ -0,0 +1,80 @@
...
+
+  fetchStats();
+
+  function fetchStats() {
+    $scope.loading = true;
+
+    return api({
+      url: '/platformManagement/dbStats'
+    }).then(function(response) {
+      if (response.data) {
```
AlphaHydrae: Superfluous `if`: `response.data` will always pass this condition even if empty.

client/widgets/db-stats/stats.directive.js
```diff
@@ -0,0 +1,80 @@
...
+        });
+
+        $scope.total.rowsCountTrend = totalCumulativeTrends;
+
+        // Update the view
+        $scope.stats = stats;
+        $scope.loading = false;
+      }
+
+      $scope.started = true;
```
AlphaHydrae: Seems to be unused.

client/widgets/db-stats/stats.template.html.slim
```diff
@@ -0,0 +1,56 @@
+.db-stats.panel.panel-default
+  .panel-heading
+    help-button
+      p Help will be there
+    h3.panel-title DB table metrics
+  .panel-body
+    .row ng-if='loading && stats.length === 0'
```
AlphaHydrae: This is never displayed, as `$scope.stats` is only set once the request is complete, therefore `$scope.stats.length` is undefined.

client/widgets/db-stats/stats.template.html.slim
```diff
@@ -0,0 +1,56 @@
...
+      .col-md-12
+        p Loading...
```
AlphaHydrae: Add the `.loading` class.

client/widgets/db-stats/stats.template.html.slim
```diff
@@ -0,0 +1,56 @@
...
+    .row.row-labels ng-if='stats.length > 0'
+      .col-md-12
+        table.table
+          thead.hidden-md.hidden-sm.hidden-xs
```
AlphaHydrae: FYI `.visible-lg` is equivalent to `.hidden-md.hidden-sm.hidden-xs`.

client/widgets/org-top-stats/stats.template.html.slim
```diff
@@ -0,0 +1,65 @@
...
+              td.projects.text-right.db-stats-right-border.hidden-md.hidden-sm.hidden-xs {{ stat.projectsCount | number }}
+              td.tests.text-right.db-stats-right-border {{ stat.testsCount | number }}
+              td.payloads.text-right.db-stats-right-border {{ stat.payloadsCount | number }}
+              td.results-count.text-right {{ stat.resultsCount | number }}
+              td.results-prop.text-right.hidden-md.hidden-sm.hidden-xs {{ stat.resultsProp | number:2 }}%
+              td.results-trend.text-center
+                sparkline ng-if='stat.resultsTrend' data='stat.resultsTrend'
+          tfoot
+            tr
+              td.name.db-stats-right-border Top total
```
AlphaHydrae: The meaning would be clearer with "Top 5 total". Also, I would expect this `td` and the next `tr`'s first `td` to be `th` elements instead. It will make it visually clearer that they are totals.

client/widgets/org-stats/stats.directive.js
```diff
@@ -0,0 +1,89 @@
...
+    scope: {
+      organization: '='
+    }
+  };
+}).controller('OrgStatsWidgetCtrl', function(api, $scope, orgs) {
+  orgs.addAuthFunctions($scope);
+
+  _.defaults($scope, {
+    params: {
+      organizationId: $scope.organization.id
```
AlphaHydrae: This can produce an NPE when no organization is pre-selected (`$scope.organization` can be null).

client/widgets/org-stats/stats.template.html.slim
```diff
@@ -0,0 +1,49 @@
+.org-stats.panel.panel-default
+  .panel-heading
+    help-button
+      p Help will be there
+    h3.panel-title Organization DB stats
+  .panel-body
```
AlphaHydrae: This widget currently displays nothing when no organization is pre-selected. It should display the organization select component, and a message prompting the user to select and organization: "Select an organization to view its DB stats."

client/widgets/org-stats/stats.directive.js
```diff
@@ -0,0 +1,89 @@
...
+  fetchStats();
+
+  function fetchStats() {
+    $scope.loading = true;
+
+    return api({
+      url: '/platformManagement/orgStats',
+      params: $scope.params
+    }).then(function(response) {
+      if (response.data) {
```
AlphaHydrae: Superfluous `if`.

AlphaHydrae:
What do you think of renaming the widgets like this?

* Top 5 largest organizations
* Organization stats
* Database size

I think those names are more user-friendly.

features/platform_management.feature
```diff
@@ -0,0 +1,252 @@
...
+        test_payloads_reports
+        test_reports
+        test_result_contributors
+        test_results
+        test_results_tickets
+        tickets
+        user_registrations
+        users
+      """
+    And the following tables contains trends:
```
AlphaHydrae: "contains" -> "should contain"

app/api/probe_dock/organizations_api.rb
Prevole: Done

client/widgets/db-stats/stats.directive.js
Prevole: Done

client/widgets/db-stats/stats.directive.js
Prevole: Done

client/widgets/db-stats/stats.directive.js
Prevole: Done

client/widgets/db-stats/stats.template.html.slim
Prevole: Done

client/widgets/db-stats/stats.template.html.slim
Prevole: Done

client/widgets/db-stats/stats.template.html.slim
Prevole: Not applicable to tables.

client/widgets/org-top-stats/stats.template.html.slim
Prevole: Fixed

client/widgets/org-stats/stats.directive.js
Prevole: Done

client/widgets/org-stats/stats.template.html.slim
Prevole: Done

client/widgets/org-stats/stats.directive.js
Prevole: Done

features/organization.feature
```diff
@@ -0,0 +1,96 @@
...
+        "createdAt": "@iso8601",
+        "updatedAt": "@iso8601"
+      }]
+      """
+    And nothing should have been added or deleted
+
+
+
+  @authorization
+  Scenario: A anonymous user should not be able to retrieve any organization when administered filter is set.
```
AlphaHydrae: **An** anonymous user...

features/organization.feature
```diff
@@ -0,0 +1,96 @@
...
+        "membershipsCount": 1,
+        "createdAt": "@iso8601",
+        "updatedAt": "@iso8601"
+      }]
+      """
+    And nothing should have been added or deleted
+
+
+
+  Scenario: A super admin should be able all the organization administered or not.
```
AlphaHydrae: In many other tests I have used the terminology "Probe Dock admin" instead of "super admin". We should make sure we use the same everywhere.

features/platform_management.feature
```diff
@@ -0,0 +1,252 @@
...
+    - Tests
+    - Results
+
+
+
+  Scenario: A Probe Dock administrator should retrieve the DB stats
+    Given user palpatine who is a Probe Dock admin exists
+    When palpatine sends a GET request to /api/platformManagement/dbStats
+    Then the response code should be 200
+    And the response body should be a JSON array with the following tables:
```
AlphaHydrae: I find the naming of this assertion confusing. It starts with "And the response body should be a JSON array...", leading me to believe that I'm going to see JSON in the text block below, but then it says "... with the following tables", and then I understand because we've discussed about it but I don't think anyone else would.

I think something like "And the response should contain DB stats for the following tables:" would be clearer.

PR #51, opened 2016-03-22
Results history UI improvement

Prevole:


client/widgets/test-results/results.directive.js
```diff
@@ -96,7 +95,7 @@ angular.module('probedock.testResultsWidget').directive('testResultsWidget', fun
         $scope.pagination = response.pagination();
```
AlphaHydrae: Superfluous `if (response.data)` one line above.

PR #54, opened 2016-03-29
Non unique technical user

Prevole:


app/models/user.rb
```diff
@@ -48,7 +49,9 @@ class User < ActiveRecord::Base
 
   strip_attributes except: :password_digest
   # TODO: validate min name length
-  validates :name, presence: true, uniqueness: true, length: { maximum: 25, allow_blank: true }, format: { with: /\A[a-z0-9]+(?:-[a-z0-9]+)*\Z/i, allow_blank: true }
+  validates :name, presence: true, length: { maximum: 25, allow_blank: true }, format: { with: /\A[a-z0-9]+(?:-[a-z0-9]+)*\Z/i, allow_blank: true }
+  validates :name, uniqueness: { unless: :technical }
+  validate :technical_user_is_unique_in_org, on: :create, if: :technical
```
AlphaHydrae: Why only validate on create? Users can be renamed.

app/models/user.rb
```diff
@@ -110,4 +113,19 @@ def complete_registration
       registration.organization.update_attribute :active, true if registration.organization.present?
     end
   end
+
+  def technical_user_is_unique_in_org
+    # This validation make the assumption a technical user has only and only one membership at the creation
+    if User.joins(:organizations).where(name: name, 'organizations.id': memberships.first.organization.id).count > 0
```
AlphaHydrae: Due to user renaming being possible for both human and technical users, I think the validations should be rewritten to allow creating a human and a technical user with the same name. They will be much simpler that way. It also corresponds to the unique database constraint on the normalized name.

This validation should have `technical: true` added to the `where` clause.

The `validates :name, uniqueness: { unless: : technical }` validation on line 53 should be rewritten as a method, to check only human users.

app/models/user.rb
```diff
@@ -110,4 +113,19 @@ def complete_registration
...
+      errors.add :name, :must_be_unique_by_org
+    end
+  end
+
+  def normalize_name
+    self.normalized_name = if self.human?
```
AlphaHydrae: The second `self` is superfluous. (The first is needed because otherwise it would be a variable assignment.)

client/validations/uniqueTechnicalUserName.directive.js
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        var userId = $scope.user ? $scope.user.id : null;
+
+        return api({
+          url: '/users',
+          params: {
+            name: modelValue,
+            organizationId: $scope.currentOrganization.id,
+            pageSize: 1
```
AlphaHydrae: Limit query to technical users.

**ALSO** modify the `uniqueUserName` directory to only query human users.

db/migrate/20160323133253_remove_unique_user_constraint.rb
```diff
@@ -0,0 +1,22 @@
...
+  def up
+    remove_index :users, :name
+
+    # This column require the space for <human|technical>[-<organizationNormalizedName>]-<name.downcase>
+    add_column :users, :normalized_name, :string, null: true, limit: 100
+
+    add_index :users, :normalized_name, unique: true
+
+    User.all.each do |user|
+      user.save
```
AlphaHydrae: Avoid doing this in the future unless there's no other choice. Always prefer `update_all`.

It's dangerous because the implementation can change between the time the migration is written and the time it is run. It also updates all the users' timestamps, which I don't think we should do in cases like this.

We won't re-run it on trial, but you can rewrite it for demo.

```
User.all.each do |user|
  normalized_name = ... # hard-code implementation here
  User.where(id: user.id).update_all(normalized_name: normalized_name)
end
```

spec/models/user_spec.rb
```diff
@@ -161,13 +162,13 @@
...
+      user = build(:user)
       user.emails = []
       expect(user).not_to be_valid
     end
 
-    describe "with an existing user" do
-      let!(:user){ create :user }
+    describe 'with an existing user' do
+      let!(:user){ create(:user) }
       it(nil, probedock: { key: 'f9be952c7792' }){ should validate_uniqueness_of(:name) }
```
AlphaHydrae: Add a test to ensure that a human user can be created if a technical user exists with the same name.

spec/models/user_spec.rb
```diff
@@ -177,19 +178,55 @@
...
+        create(:technical_user, name: 'tech', organization: organization)
+        second_tech_user = build(:technical_user, name: 'tech', organization: organization)
+        second_tech_user.memberships << build(:membership, user: second_tech_user, organization: organization)
+        second_tech_user
+      end
+      it(nil, probedock: { key: 'iwzr' }) do
+        expect(subject).not_to be_valid
+        expect(subject.errors[:name].size).to eq(1)
+        expect(subject.errors[:name][0]).to eq('must be unique in organization')
+      end
```
AlphaHydrae: Add a test to ensure that a technical user can be created in an organization with a human member who has the same name.

spec/models/user_spec.rb
```diff
@@ -177,19 +178,55 @@
       end
     end
 
-    describe "for inactive users" do
-      subject{ described_class.new active: false }
+    describe 'with an existing technical user' do
+      subject do
+        organization = create(:organization)
+        create(:technical_user, name: 'tech', organization: organization)
```
AlphaHydrae: I prefer declaring such data in `let` blocks, even if I only use it in the `subject`, because it makes the `subject` less complex and easier to read and it makes things potentially easier when you add future tests (which I suggested doing in the previous comment).

app/models/user.rb
```diff
@@ -110,4 +113,19 @@ def complete_registration
...
+      "human-#{name.downcase}"
+    else
+      "technical-#{memberships.first.organization.normalized_name}-#{name.downcase}"
```
AlphaHydrae: This implementation may cause collisions in some cases. I created a separate issue for this: https://probe.storiesonboard.com/m/probe-dock-roadmap/!card/curfjHen8bIdbDnXJqDZ2w

spec/models/user_spec.rb
```diff
@@ -168,24 +168,99 @@
...
+      end
+
+      let(:human_user) do
+        human_user = build(:user, name: 'anothername', organization: organization)
+        human_user.memberships << build(:membership, user: human_user, organization: organization)
+        human_user.save
+        human_user
+      end
+
+      it 'possible when the human user is renamed with the same name of an existing technical user', probedock: { key: 'wofz' } do
```
AlphaHydrae: Add "is" at the beginning of this `it` and the next, and remove it from the parent `describe`.

PR #56, opened 2016-03-30
Reports category filter fix

Prevole:


AlphaHydrae:
There should be a regression test for the API modification.

AlphaHydrae:
The join currently goes through the reports table. We added a cache of the categories on the payloads, it should be used instead.

Be careful not to make two joins on the test payloads. These two lines should be modified: https://github.com/probedock/probedock/blob/master/app/api/probe_dock/reports_api.rb#L81-L82

Prevole:
Ready

features/support/cucumber_spec_helper.rb
```diff
@@ -27,6 +27,11 @@ def cucumber_request_data data_type, doc
     if data_type.match /^(xml|json)$/i
       type = data_type.downcase.to_sym
       content_type = "application/#{type}"
+    else
+      matched_type = data_type.match /.*\+(xml|json)$/i
```
AlphaHydrae: Missing parentheses. Please also rewrite the regexp as `/(?:^application\/|.*\+)(xml|json)$/i` so that it matches `application/json` and `application/xml` as well.

PR #58, opened 2016-04-04
Reports filtering status

Prevole:


client/pages/report-list/list.css.styl
```diff
@@ -4,13 +4,34 @@
     margin-top 15px
     padding 15px
 
+    .select-status
+      align-items center
+      flex-wrap wrap
+      display flex
+
+      .select-label
+        font-weight bold
```
AlphaHydrae: Any reason why this long sentence should be in bold style? I think it looks much better without it.

app/api/probe_dock/reports_api.rb
```diff
@@ -116,6 +122,31 @@ def serialization_options reports
...
+
+            complete_clause = ''
+            where_clauses.each_with_index do |clause, idx|
+              complete_clause += clause
+              if idx < where_clauses.size - 1
+                complete_clause += ' OR '
+              end
+            end
+
+            paginated_rel = paginated_rel.where(complete_clause)
```
AlphaHydrae: Using `where_clauses.join(' OR ')` here would allow you to remove the entire previous code block.

app/api/probe_dock/reports_api.rb
```diff
@@ -116,6 +122,31 @@ def serialization_options reports
             paginated_rel = paginated_rel.where('categories.name in (?)', params[:categoryNames].collect(&:to_s).to_a)
           end
 
+          if array_param?(:status) || array_param?(:kind)
+            where_clauses = []
```
AlphaHydrae: I would name this variable `payload_conditions`.

app/api/probe_dock/reports_api.rb
```diff
@@ -116,6 +122,31 @@ def serialization_options reports
             paginated_rel = paginated_rel.where('categories.name in (?)', params[:categoryNames].collect(&:to_s).to_a)
           end
 
+          if array_param?(:status) || array_param?(:kind)
```
AlphaHydrae: This should be split into two `if` blocks. It currently is bugged if you combine status and kind parameters.

If you search for `passed` status and `new` kind, you might get payloads that do not contain any passed tests but that do contain new tests. This is inconsistent with the overall API behavior where separate parameters are always "AND-ed".

app/api/probe_dock/reports_api.rb
```diff
@@ -116,6 +122,31 @@ def serialization_options reports
...
+            where_clauses = []
+
+            if array_param?(:status)
+              where_clauses << 'test_payloads.passed_results_count > 0' if params[:status].include?('passed')
```
AlphaHydrae: This condition should be `test_payloads.passed_results_count > test_payloads.inactive_passed_results_count` because `passed_results_count` includes inactive results.

app/api/probe_dock/reports_api.rb
```diff
@@ -116,6 +122,31 @@ def serialization_options reports
...
+              where_clauses << 'test_payloads.results_count - test_payloads.passed_results_count - test_payloads.inactive_results_count > 0' if params[:status].include?('failed')
+              where_clauses << 'test_payloads.inactive_results_count > 0' if params[:status].include?('inactive')
+            end
+
+            if array_param?(:kind)
```
AlphaHydrae: I suggest that we change this to a `newTests` boolean parameter that can be either true (retrieve payloads with new tests) or false (retrieve payloads with no new tests, i.e. only existing tests).

AlphaHydrae: Addendum: there will therefore be no longer be a need for an OR query for this particular parameter.

client/pages/report-list/list.controller.js
```diff
@@ -21,6 +21,37 @@ angular.module('probedock.reportListPage').controller('ReportListPageCtrl', func
...
+    }
+  }
+
+  $scope.$watch('statuses', function() {
+    manageToggleParams('statuses', 'status');
+  }, true);
+
+  $scope.$watch('kinds', function() {
+    manageToggleParams('kinds', 'kind');
+  }, true);
```
AlphaHydrae: Remove legacy code (from `manageToggleParams` function to here).

client/pages/report-list/list.template.html.slim
```diff
@@ -17,6 +17,21 @@
                   project-version-select prefix='report-list' organization='currentOrganization' model-object='reportsList.params' model-property='projectVersionNames' multiple='true' unique-by='name' extract='name'
                 .col-lg-3.col-md-6
                   category-select prefix='report-list' organization='currentOrganization' model-object='reportsList.params' multiple='true'
+              .row
+                .col-md-12
+                  .select-status
+                    .select-label
+                      | Show the reports containing at least one
+                      ui-select id='reports-list-status' ng-model='selectParams.status' reset-search-input=true
```
AlphaHydrae: Use `ui-select<>` for spacing (also for the next one).

PR #59, opened 2016-04-04
Added repo URL validation and repo URL pattern

Prevole:


features/step_definitions/project_steps.rb
```diff
@@ -7,5 +7,11 @@
   options[:display_name] = name if name != options[:name]
   options[:repo_url] = repo_url if repo_url
 
-  add_named_record name, create(:project, options)
+  add_named_record(name, create(:project, options))
 end
+
+Given /^the project (.+) updated with the repo url pattern (.+)$/ do |name,repo_url_pattern|
```
AlphaHydrae: To be more consistent with the way we have named such steps before, I would rephrase it to: "project A has repo url pattern B".

PR #68, opened 2016-04-19
Required fields

Prevole:


AlphaHydrae:
Additional changes:

* Use `required: true` for the directive. That way there's no longer an intermediate `label-required` element in the DOM, just in case Bootstrap decides to change the way the CSS for `label` elements is defined (for example `.form-group > label` instead of `.form-group label`).
* Use `ng-attr-for="{{ for }}"` instead of `for="{{ for }}"`. This has the advantage of not putting the `for` attribute in the DOM when the value is not present.

client/pages/member-edit-modal/modal.template.html.slim
```diff
@@ -5,7 +5,7 @@ form novalidate=true name='membershipForm' ng-submit='membershipForm.$valid && s
     h3.modal-title ng-if='!editedMembership.id' New Member
   .modal-body
     .form-group ng-if='!membership.id'
-      label Type
+      label-required for='technical' Type
```
AlphaHydrae: I wouldn't put "required" next to a radio button that cannot be unselected anyway.

PR #69, opened 2016-04-19
User friendly names in place of tech/human names

Prevole:


client/pages/org-edit-modal/modal.template.html.slim
```diff
@@ -4,19 +4,13 @@ form#org-edit-modal novalidate=true name='orgForm' ng-submit='orgForm.$valid &&
     h3.modal-title ng-if='editedOrg.id' Edit {{ organization | orgName }}
     h3.modal-title ng-if='!editedOrg.id' New Organization
   .modal-body
-    .form-group ng-class='{"has-error": orgForm.name.$dirty && orgForm.name.$invalid}'
-      label for='orgName' Name
-      input#orgName.form-control type='string' name='name' placeholder='Name (used in URLs)' required=true maxlength=50 ng-pattern='/^[a-z0-9]+(?:-[a-z0-9]+)*$/i' unique-org-name=true autofocus=true ng-model='editedOrg.name' ng-model-options='{updateOn: "default blur", debounce: {default: 500, blur: 0}}'
+    .form-group ng-class='{"has-error": orgForm.displayName.$dirty && orgForm.displayName.$invalid}'
+      label for='orgDisplayName' Name
+      input#orgDisplayName.form-control type='string' name='displayName' maxlength=50 ng-model='editedOrg.displayName' unique-org-name=true
```
AlphaHydrae: You should not have removed `ng-model-options='{updateOn: "default blur", debounce: {default: 500, blur: 0}}'`.

client/pages/project-edit-modal/modal.template.html.slim
```diff
@@ -3,22 +3,16 @@ form#project-edit-modal novalidate=true role='form' name='projectForm' ng-submit
...
   .modal-body
-    .form-group ng-class='{"has-error": projectForm.name.$dirty && projectForm.name.$invalid}'
-      label for='projectName'
+    .form-group ng-class='{"has-error": projectForm.displayName.$dirty && projectForm.displayName.$invalid}'
+      label for='projectDisplayName'
         | Name
         span.required<
           | (required)
-      input#projectName.form-control type='string' name='name' placeholder='Name (used in URLs)' required=true maxlength=50 ng-pattern='/^[a-z0-9]+(?:-[a-z0-9]+)*$/i' unique-project-name=true autofocus=true ng-model='editedProject.name' ng-model-options='{updateOn: "default blur", debounce: {default: 500, blur: 0}}'
+      input#projectDisplayName.form-control type='string' name='displayName' placeholder='Name' maxlength=50 ng-model='editedProject.displayName' unique-project-name=true autofocus=true
```
AlphaHydrae: You should not have removed `ng-model-options='{updateOn: "default blur", debounce: {default: 500, blur: 0}}'`.

client/pages/org-edit-modal/modal.template.html.slim
```diff
@@ -4,19 +4,13 @@ form#org-edit-modal novalidate=true name='orgForm' ng-submit='orgForm.$valid &&
     h3.modal-title ng-if='editedOrg.id' Edit {{ organization | orgName }}
     h3.modal-title ng-if='!editedOrg.id' New Organization
   .modal-body
-    .form-group ng-class='{"has-error": orgForm.name.$dirty && orgForm.name.$invalid}'
-      label for='orgName' Name
-      input#orgName.form-control type='string' name='name' placeholder='Name (used in URLs)' required=true maxlength=50 ng-pattern='/^[a-z0-9]+(?:-[a-z0-9]+)*$/i' unique-org-name=true autofocus=true ng-model='editedOrg.name' ng-model-options='{updateOn: "default blur", debounce: {default: 500, blur: 0}}'
+    .form-group ng-class='{"has-error": orgForm.displayName.$dirty && orgForm.displayName.$invalid}'
+      label for='orgDisplayName' Name
```
AlphaHydrae: We agreed (a while ago) that the naming convention for new HTML classes and IDs would be kebab-case, i.e. `org-display-name`, like Bootstrap does.

client/pages/org-edit-modal/modal.template.html.slim
```diff
@@ -4,19 +4,13 @@ form#org-edit-modal novalidate=true name='orgForm' ng-submit='orgForm.$valid &&
...
+      input#orgDisplayName.form-control type='string' name='displayName' maxlength=50 ng-model='editedOrg.displayName' unique-org-name=true
       p.help-block
-        | This is the identifier of the organization and is used in URLs.
-        |  It can only contain alphanumeric characters and single hyphens (not at the beginning or end).
-      .alert.alert-danger ng-if='orgForm.name.$error.uniqueOrgName'
+        | This is the name of the organization as shown in Probe Dock.
```
AlphaHydrae: This message no longer makes sense. The "as shown in Probe Dock" was meant to differentiate it from the identifier. I would replace the full message with "The unique name of the organization, up to 50 characters."

client/services/api/api.service.js
```diff
@@ -100,5 +100,15 @@ angular.module('probedock.api').factory('api', function(apiPagination, auth, $ht
     }
   };
 
+  api.slugify = function(str) {
+    return str
+      .replace(/[^a-z0-9\- ]+/gi, '')
+      .replace(/ +/g, '-')
+      .replace(/\-+/g, '-')
+      .replace(/\-+$/, '')
+      .replace(/^\-+/, '')
```
AlphaHydrae: As discussed, remove spaces instead of replacing them with hyphens. That will avoid most collisions.

I would replace the whole implementation of this function with:

```
return str.replace(/[^a-z0-9]+/gi, '').toLowerCase()
```

This will keep letters and digits, and nothing else.

AlphaHydrae:
As discussed, the way we decided to implement this causes a few problems with organizations that had no display name. The display name should be made mandatory on the server side. The migration will also need to set it to the name when it is missing. This concerns organizations and projects.

client/pages/project-edit-modal/modal.template.html.slim
```diff
@@ -3,22 +3,16 @@ form#project-edit-modal novalidate=true role='form' name='projectForm' ng-submit
...
       p.help-block
-        | This is the identifier of the project and is used in URLs.
-        |  It can only contain alphanumeric characters and single hyphens (not at the beginning or end).
-      .alert.alert-danger ng-if='projectForm.name.$error.uniqueProjectName'
+        | This is the name of the project as shown in ProbeDock.
```
AlphaHydrae: Replace with "The name of the project, unique within the organization, up to 50 characters."

client/pages/user-registration/registration.controller.js
```diff
@@ -25,14 +25,7 @@ angular.module('probedock.userRegistrationPage').controller('UserRegistrationPag
...
     value = value || '';
-
-    $scope.organizationNamePlaceholder = value
-      .replace(/[^a-z0-9\- ]+/gi, '')
-      .replace(/ +/g, '-')
-      .replace(/\-+/g, '-')
-      .replace(/\-+$/, '')
-      .replace(/^\-+/, '')
-      .toLowerCase();
+    $scope.organizationNamePlaceholder = api.slugify(value);
```
AlphaHydrae: This change to the registration form implementation is not sufficient. This is the code that sets the placeholder in the `name` field, computed from the `displayName` field, but what we actually want to do is remove the name field entirely, so this code should go as well.

Actually, [most of the code](https://github.com/probedock/probedock/blob/friendly-names/client/pages/user-registration/registration.controller.js#L26-L57) in the controller is obsolete if we only have the `displayName` field. [This part](https://github.com/probedock/probedock/blob/friendly-names/client/pages/user-registration/registration.controller.js#L13-L15) should be replaced with code similar to what you introduced in the model services.

The template should be updated to have only the `displayName` field, and to use the validation. Also, the organization part of the template currently has the display name and name on one line, and the public checkbox on another line. Since we're removing the name field, you can put the display name and public checkbox on the same line (with `.col-sm-6` divs).

client/pages/user-registration/registration.template.html.slim
```diff
@@ -6,7 +6,7 @@
   .registration-successful ng-if='registration'
     p.lead.text-success Registration successful!
     p.text-success
-      | We have sent a confirmation e-mail to 
+      | We have sent a confirmation e-mail to
       strong {{ registration.user.email }}
```
AlphaHydrae: Add a `<` to this `strong` tag to replace the trailing space you removed.

client/pages/project-edit-modal/modal.template.html.slim
```diff
@@ -8,9 +8,9 @@ form#project-edit-modal novalidate=true role='form' name='projectForm' ng-submit
         | Name
         span.required<
           | (required)
-      input#projectDisplayName.form-control type='string' name='displayName' placeholder='Name' maxlength=50 ng-model='editedProject.displayName' unique-project-name=true autofocus=true
+      input#projectDisplayName.form-control type='string' name='displayName' placeholder='Name' maxlength=50 ng-model='editedProject.displayName' unique-project-name=true autofocus=true ng-model-options='{updateOn: "default blur", debounce: {default: 500, blur: 0}}'
       p.help-block
-        | This is the name of the project as shown in ProbeDock.
+        | The unique name of the project, up to 50 characters.
```
AlphaHydrae: "The name of the project, unique within the organization, up to 50 characters."

client/pages/user-registration/registration.template.html.slim
```diff
@@ -51,22 +51,14 @@
...
-        label for='organization-display-name' Name
-        input#organization-display-name.form-control type='string' maxlength=50 ng-model='organization.displayName' required=true
-        p.help-block
-          | The name of the organization.
-
-    .col-sm-6.col-xs-12
-      .form-group ng-class='{"has-error": registrationForm.name.$dirty && registrationForm.name.$invalid || organizationNamePlaceholderTaken}'
+    .col-sm-12.col-xs-12
+      .form-group ng-class='{"has-error": registrationForm.organizationName.$dirty && registrationForm.organizationName.$invalid}'
         label for='organization-name' Identifier
```
AlphaHydrae: Replace "Identifier" with "Name".

client/pages/user-registration/registration.template.html.slim
```diff
@@ -51,22 +51,14 @@
...
-    .col-sm-6.col-xs-12
-      .form-group
-        label for='organization-display-name' Name
-        input#organization-display-name.form-control type='string' maxlength=50 ng-model='organization.displayName' required=true
-        p.help-block
-          | The name of the organization.
-
-    .col-sm-6.col-xs-12
-      .form-group ng-class='{"has-error": registrationForm.name.$dirty && registrationForm.name.$invalid || organizationNamePlaceholderTaken}'
+    .col-sm-12.col-xs-12
```
AlphaHydrae: See what it looks like with a `.col-sm-6` here and for the checkbox. Keep your preferred version (I don't care which).

Either use two `.col-sm-6.col-xs-12` or two `.col-xs-12`, but `.col-sm-12.col-xs-12` is redundant.

PR #70, opened 2016-04-22
Client upgrade

AlphaHydrae:


AlphaHydrae:
The following changes are worth reviewing and testing carefully:

* `.travis.yml`
  * Temporary fix due to an incompatibility between Angular 1.5.x and PhantomJS 1.9.x which is provided by Travis. We manually install PhantomJS 2 to solve the build errors.
* `Gruntfile.js`
  * Added a processing function to the copy task to remove source map URLs which were causing errors in the console.
* `client/pages/org-edit-modal/modal.service.js`
  * The service has been updated to automatically load the organization if given only the name or ID, like the corresponding project modal does.
* `client/pages/report-list/list.controller.js`
  * Due to changes in UI Bootstrap's tabs, the `ReportListTabCtrl` controller has been deleted and integrated into `ReportListPageCtrl`.
  * The way tabs are selected has also been updated.
* `client/pages/user-list/list.controller.js`
  * The way tabs are selected has been updated.
* `client/routes.js`
  * `angular-ui-router-title` has been removed due to incompatibility with Angular UI Router 1.x. Its functionality has been integrated into this file.
  * Angular UI Router's `$stateParams` no longer includes the parameters of parent states. Resolve functions must be used instead (see the new `routeOrgName` and `routeUserId` resolve functions). All `$title` resolve functions have been updated to work with the new resolve mechanisms.
* `client/services/bootstrap/bootstrap.directive.js`
  * The dirty fix to detect resizing has been removed since it should now work with Angular 1.5.x.
* `client/services/states/states.service.js`
  * The service has been completely refactored and now plugs into Angular UI Router's new transition service.
  * Angular UI Router's `$stateChangeStart` and `$stateChangeSuccess` events are now deprecated in favor of its new transition service. All references to these events throughout the application have been replaced with callback registration on our `states` service.

client/pages/org-edit-modal/modal.service.js
```diff
@@ -1,36 +1,52 @@
...
+}).controller('OrgEditModalCtrl', function(api, forms, orgs, $scope, $uibModalInstance) {
 
-  $scope.organization = {};
+  $scope.organization = $scope.organization || {};
   $scope.editedOrg = {};
 
-  if ($stateParams.orgName) {
+  if ($scope.organization && $scope.organization.id) {
+    // Edit the specified organization.
+    reset();
```
Prevole: It is a bit confusing to have a reset function with a comment that says we edit the specified organization. It took me few minutes to understand that we set the form model to the correct organization depending the situation (we have an organization object, or an organization id meaning we have to retrieve the organization from server and then to set the form model).

I would suggest to rename the function `reset` to something like `setEditedOrganization`. We can keep `$scope.reset = <fnNewName>` as I suppose the button reset is bound to this `reset` fn.

AlphaHydrae: Done, but I prefer `resetEditedOrganization`.

client/services/states/states.service.js
```diff
@@ -1,59 +1,179 @@
...
 
-  var onStateCallbacks = [];
+  var currentState = null,
+      onStateChangeStartCallbacks = [],
+      onStateChangeSuccessCallbacks = [];
 
-  $rootScope.$on('$stateChangeSuccess', checkState);
+  var service = {
 
-  function checkState(event, toState, toParams, fromState, fromParams) {
```
Prevole: The comments on the onStateChnage* functions are really useful.

I would just suggest to add a little precision about the fact a scope is mandatory and in case we want to add a callback that will persist during the whole lifecyle of the app. In other terms, explanation on when to user $rootScope.

bower.json
```diff
@@ -9,41 +9,41 @@
...
-    "bootstrap": "~3.3.0",
-    "drop-ng": "git@github.com:stevenh77/drop-ng.git#b3f30fb275a60ae351ba7624a2c6b1d6dd78136a",
-    "dropzone": "~4.2.0",
-    "font-awesome": "~4.5.0",
-    "jquery": "~2.1.1",
-    "js-yaml": "~3.3.0",
+    "angular-ui-router": "~1.0.0",
+    "angular-ui-select": "~0.16.1",
+    "bootstrap": "~3.3.6",
+    "drop-ng": "~0.1.5",
```
Prevole: I took the opportunity to update this dependency too as there is a release now.

client/pages/dashboard/dashboard.header.controller.js
```diff
@@ -1,22 +1,17 @@
-angular.module('probedock.dashboardPage').controller('DashboardHeaderCtrl', function(orgEditModal, $scope, $state, $stateParams) {
+angular.module('probedock.dashboardPage').controller('DashboardHeaderCtrl', function(orgEditModal, $scope, $state, states) {
 
-  var modal;
-  $scope.currentState = $state.current.name;
+  states.onStateChangeSuccess($scope, 'org.dashboard.default.edit', function(state, params, resolves) {
 
-  $scope.$on('$stateChangeSuccess', function(event, toState) {
+    var modal = orgEditModal.open($scope, {
+      organizationName: resolves.routeOrgName
```
Prevole: Question: When do we need to inject routeOrgName and when do we use it from resolves ?

client/pages/layout/layout.nav.controller.js
```diff
@@ -1,20 +1,18 @@
...
-  var state = $state.current;
-  $rootScope.$on('$stateChangeSuccess', function(event, toState, toStateParams) {
-    state = toState;
-    $scope.orgName = toStateParams.orgName;
+  var currentStateName = null;
+  states.onStateChangeSuccess($scope, true, function(state, params, resolves) {
+    currentStateName = state.name;
   });
 
   $scope.baseStateIs = function() {
```
Prevole: We should take some time a day to think about that:

```
$scope.baseStateIs = function() {
  console.log('here');
  ...
}
```
Accessing the page: /

In the console:
`here` appears -> 92 times

client/pages/dashboard/dashboard.header.controller.js
AlphaHydrae: I think both would work in this case, because it seems the dashboard header controller is recreated even if you say on the dashboard and switch organizations.

But in a case where you would change state without this controller being recreated, then you would need to get it from the state's resolves and not from the original injection, which wouldn't be updated.

client/pages/layout/layout.nav.controller.js
AlphaHydrae: Any function called from the views will be called for each digest cycle. Maybe it would be faster to listen to state transitions, store a `$scope.baseState` variable, and replace the call in the view by `baseState == "help"`. But I don't even know if it would actually be any faster.

Add a story if you want to check it out.

PR #73, opened 2016-05-03
Rep filters failed fix

Prevole:
No regression tests added, in fact, there was already a test that validated that but the test was not correct.


features/reports_filtering.feature
```diff
@@ -356,6 +356,18 @@ Feature: Various filters to get reports
     Then the response should be HTTP 200 with the following JSON:
       """
       [{
+        "id": "@idOf: D",
+        "duration": "@integer",
+        "resultsCount": 1,
+        "passedResultsCount": 0,
+        "inactiveResultsCount": 1,
+        "inactivePassedResultsCount": 1,
```
AlphaHydrae: The modification to this test is incorrect. Report D has no failing tests, so it should not be returned for this query. The report counters themselves are in fact invalid. It should not be possible to have `passedResultsCount` equal 0 and `inactivePassedresultsCount` equal 1, as the first is supposed to be a superset of the other.

The bug is [here](https://github.com/probedock/probedock/blob/8ec7561ba42a50fdfe72bfc7d942d1451f8e44c8/features/step_definitions/result_steps.rb#L28) in the test data generation.

This line is incorrect:

```ruby
payload.passed_results_count += 1 if options[:passed] && options[:active]
```

`passed_results_count` should be incremented `if options[:passed]`, regardless or whether the result is active or not.

PR #76, opened 2016-05-03
Fixed health bars appearance

Prevole:


client/components/report-health-bar/bar.directive.js
```diff
@@ -5,12 +5,23 @@ angular.module('probedock.reportHealthBar').directive('reportHealthBar', functio
...
+      percentageMinimumThreshold: '@'
     }
   };
 }).controller('ReportHealthBarCtrl', function($attrs, reports, $scope) {
+  _.defaults($scope, {
+    percentageMinimumThreshold: 5
+  });
 
   $scope.percentages = reports.percentages($scope.report);
+  $scope.widthPercentages = effectiveWidths($scope.percentages);
```
AlphaHydrae: Modify existing `reports.percentages` function to return floats, not integers. And modify new algorithm as discussed. Please put an example in the comments.

PR #77, opened 2016-06-10
Reports shared filters

Prevole:


client/services/tables/tables.service.js
```diff
@@ -6,15 +6,15 @@ angular.module('probedock.tables').factory('tables', function(api) {
...
+        params: _.extend({}, options.params)
       };
 
       list.refresh = function(table) {
 
         table.pagination.start = table.pagination.start || 0;
         table.pagination.number = table.pagination.number || options.pageSize || 15;
 
-        var params = _.extend({}, options.params, list.params, {
+        var params = _.extend({}, list.params, {
```
AlphaHydrae: Is this change necessary?

AlphaHydrae:
If I start from the reports list with no filters and I then select a runner filter, a project filter, a version filter and a category filter (4 filters), it takes me 7 clicks on the browser's back button to come back to the original state, when it should only be 4.

Maybe some of the watching code does too many modifications to the search params.

client/services/tables/tables.service.js
Prevole: Yeps. I moved the options.params to line 9. Without this update, when you receive a link with filters in the query params, the filters are not filled because the params start with empty object. We need to reflect the query params in the table params. Maybe there is a different way ?

